PR #6, opened 2015-11-25
Properly escape quotes in titles and descriptions

benbalter:
@parkr I could use your :books: here, if there's a better way to do this. I've tried doing things a million ways. Here's what I've learned:

* Kramdown converts dumb quotes to smart quotes
* Smart quotes are not escaped by `escape` or its variants, but are treated as dumb quotes for purposes of parsing and validation by many parsers
* If quotes are encoded as `&quot` before being passed to Kramdown, Kramdown changes them to dumb quotes
* All software is terrible

Halp?

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -5,7 +5,16 @@ module Jekyll
 
     def render(context)
       @context = context
-      Liquid::Template.parse(template_contents).render!(payload, info).gsub(/[\n\s]{2,}/, "\n")
+      output = Liquid::Template.parse(template_contents).render!(payload, info)
+
+      # Minify
+      output.gsub!(/[\n\s]{2,}/, "\n")
+
+      # Encode smart quotes. See https://github.com/benbalter/jekyll-seo-tag/pull/6
+      output.gsub!("\u201c", "&ldquo;")
+      output.gsub!("\u201d", "&rdquo;")
+
+      output
     end
 
     private
--- a/lib/template.html
+++ b/lib/template.html
@@ -18,7 +18,7 @@
 	{% endif %}
 {% endif %}
 {% if seo_title %}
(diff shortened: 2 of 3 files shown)


parkr:
> Halp?

Looks like the `escape` and `replace` method you have here works? Do you have a failing test I can try? You can always register new filters in a plugin if need be :smile: 

> Smart quotes are not escaped by `escape` or its variants

Interesting – I would have expected `xml_escape` to work, but alas, it uses `CGI.escapeHTML`. Try `cgi_escape`:

```ruby
>> require 'cgi'
=> true
>> CGI.escape("”") # aka cgi_escape in Jekyll
=> "%E2%80%9D"
>> CGI.escapeHTML("”") # aka xml_escape in Jekyll
=> "”"
```

benbalter:
> Looks like the escape and replace method you have here works? 

The tests pass with the code currently on the branch ~~(but it's a bit ugly IMHO). ~~

> Do you have a failing test I can try? 

Yeah, see the two previously failing tests in the PR: 

https://github.com/benbalter/jekyll-seo-tag/pull/6/files#diff-06fc5c0a8b54aaf0f8ffdb1d68786f20R38

> Try cgi_escape:

Double escapes the & that Kramdown already escaped. What we really need is `cgi_escape_once`

>  would have expected xml_escape to work, but alas, it uses CGI.escapeHTML

Yeah. [`xml_escape`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/filters.rb#L108-L110) is identical to [`escape`](https://github.com/Shopify/liquid/blob/c38eec0293d02b10427248ebb579cec2fe6788cd/lib/liquid/standardfilters.rb#L35-L37).

> You can always register new filters in a plugin if need be

Just tacked two gsubs on the end of the output to blindly escape all smart quotes (since there's no risk of double escaping).

benbalter:
@parkr Thanks for helping me think through things!

PR #12, opened 2016-01-05
Do not build Gem on CI

pathawks:
I don't think there is a reason to build the Gem from CI, is there?
That's what `script/release` is for, right?

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -3,4 +3,3 @@
 set -ex
 
 bundle exec rake spec
-bundle exec gem build jekyll-seo-tag.gemspec


benbalter:
I build as a habit (and `.gitignore` the built gem) because I've had issues where I've bumped a gem dependency and released a gem with irreconcilable dependencies. This ensures the darn thing actually builds as a sanity check. Is there a reason not to?

pathawks:
> This ensures the darn thing actually builds as a sanity check.

That makes sense. :+1:

PR #14, opened 2016-01-05
Strip trailing /index.html from seo_url

### LINKED ISSUE #2: Strip trailing /index.html from seo_url

pathawks:
Fixes #2 

This replace technique is the same as [what is used by **jekyll-sitemap**](https://github.com/jekyll/jekyll-sitemap/blob/v0.9.0/lib/sitemap.xml#L16)

What happens if the post is actually just in a subdirectory called `/index.html/`? Every thing falls apart, I guess :egg:

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -41,8 +41,8 @@
 {% endif %}
 
 {% if seo_url %}
-  <link rel="canonical" href="{{ seo_url }}{{ page.url }}" itemprop="url" />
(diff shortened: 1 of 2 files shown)


benbalter:
LGTM. Thanks, once again, @pathawks! :turtle: :cactus: :hamster: 

> What happens if the post is actually just in a subdirectory called /index.html/? Every thing falls apart, I guess

Yeah. Don't do that. I'm fine calling that a crazy edge case not worth supporting.

PR #15, opened 2016-01-06
Use baseurl to build seo_url

pathawks:


benbalter:
Nice catch. It looks like you've got a merge conflict with one of your previous PRs (thanks again), but should be ready to :ship: as soon as that's resolved.

PR #16, opened 2016-01-06
Combine two gsubs into one

pathawks:


benbalter:
That's some crazy dark-magic RegEx there. Thanks @pathawks! :ghost: :tada: :zap:

PR #19, opened 2016-01-06
Add smartify filter

### LINKED ISSUE #17: Poor Man’s Smartypants
```text
{% assign seo_title = seo_title | markdownify | strip_html | strip_newlines | escape_once %}
```

What we are doing here is rolling our own SmartyPants implementation in Liquid, which is not great.

Redcarpet has a [fantastic SmartyPants implementation](https://github.com/vmg/redcarpet#also-now-our-pants-are-much-smarter), and is even [already included in GitHub Pages](https://github.com/github/pages-gem/blob/v40/lib/github-pages.rb#L18).

We could implement our own Liquid filter (visible only to this plugin) that will call `Redcarpet::Render::SmartyPants.render`. This way, we could do all of this in one filter instead of four. Additionally, this will properly output the title `#Hashtag` instead of (currently) changing it to `Hashtag` (because Markdown treats the `#` as an `<h1>` tag, which is then stripped).

The only reason I open this as an issue instead of a PR is because requiring Redcarpet gives me pause.

pathawks:
**kramdown** Is already required by both Jekyll 2 and 3, so this is not introducing a new dependency.

This creates a stripped down Kramdown Parser that is only concerned with quotes and typographic characters. Since it is natively encoding the smart quotes for us, there is no need for the `HTML_ESCAPE_REGEX` anymore.

The `smartify` filter is only available within our template, it is not available to regular Jekyll layouts.

Fixes #17

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -1,21 +1,13 @@
+require "seo/filters.rb"
+
 module Jekyll
   class SeoTag < Liquid::Tag
 
     attr_accessor :context
 
-    HTML_ESCAPE = {
-      "\u201c".freeze => '&ldquo;'.freeze,
-      "\u201d".freeze => '&rdquo;'.freeze
-    }
-    HTML_ESCAPE_REGEX = Regexp.union(HTML_ESCAPE.keys).freeze
-
     def render(context)
       @context = context
       output = Liquid::Template.parse(template_contents).render!(payload, info)
-
-      # Encode smart quotes. See https://github.com/benbalter/jekyll-seo-tag/pull/6
-      output.gsub!(HTML_ESCAPE_REGEX, HTML_ESCAPE)
-
       output
     end
 
@@ -31,7 +23,7 @@ module Jekyll
     def info
       {
         :registers => context.registers,
-        :filters   => [Jekyll::Filters]
+        :filters   => [Jekyll::Filters, Jekyll::Seo::Filters]
       }
     end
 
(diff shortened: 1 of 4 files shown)


lib/jekyll-seo-tag.rb
```diff
@@ -34,7 +27,7 @@ def payload
     def info
       {
         :registers => context.registers,
-        :filters   => [Jekyll::Filters]
+        :filters   => [Jekyll::Filters, Jekyll::Seo::Filters]
```
parkr: I wonder if we can piggy-back on Liquid's registered filters.

pathawks: I am trying to keep the filter only available for use by the plugin, rather than using `Liquid::Template.register_filter` to make it available everywhere.

Is there a better way to go about this?

parkr: That's fair. We could also add it to Jekyll in the future of that's of interest. People seem to like smart quotes.

pathawks: I would love to see this in Jekyll :heart_eyes_cat:

parkr: I'd welcome a PR!

benbalter:
Although the current implementation is ugly, I'd rather add the smartify filter to Jekyll core for two reasons:

1. I think it's generally applicable... I know I'd use it outside of the plugin on my own site
2. If we can keep these plugins as close to liquid template + bare logic to render, it's *significantly* easier to get the code into GitHub Pages with minimal code review from a security perspective.

pathawks:
> I'd rather add the smartify filter to Jekyll core

Yes, I totally understand.

jekyll/jekyll#4323

pathawks:
> I'd rather add the smartify filter to Jekyll core

Your wish has been granted :+1:

PR #20, opened 2016-01-06
Minify template just once

pathawks:
Instead of minifying the results each time the tag is rendered, minify the template when it is read from the file and never again.


benbalter:
Huh. Never thought of that. Makes sense. :+1:

parkr:
:tada:

PR #21, opened 2016-01-10
Do not escape smart quotes

pathawks:
We should not be escaping special characters here. If the user has configured their Markdown processor to output Unicode characters, we should respect that. If a user wants to output special characters encoded as HTML entities, chances are they want that consistently across the site, not just in the `seo` tag.

I could add some tests to check if the output contains the special characters if so configured, if that would be useful.

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -3,18 +3,9 @@ module Jekyll
 
     attr_accessor :context
 
-    HTML_ESCAPE = {
-      "\u201c".freeze => '&ldquo;'.freeze,
-      "\u201d".freeze => '&rdquo;'.freeze
-    }
-    HTML_ESCAPE_REGEX = Regexp.union(HTML_ESCAPE.keys).freeze
-
     def render(context)
       @context = context
       output = Liquid::Template.parse(template_contents).render!(payload, info)
-
-      # Encode smart quotes. See https://github.com/benbalter/jekyll-seo-tag/pull/6
-      output.gsub!(HTML_ESCAPE_REGEX, HTML_ESCAPE)
 
       output
     end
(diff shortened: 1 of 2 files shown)


benbalter:
That was added via https://github.com/benbalter/jekyll-seo-tag/pull/6. It'd be awesome if the plugin could work out of the box without requiring any additional configuration. 

The issue is, if using Kramdown (the default), and we markdownify a title or description with quotes, the quotes are converted to smart quotes, which (I don't fully understand why), breaks validation when contained within an HTML attribution which itself, is wrapped in quotes. Glad to find a cleaner/more creative solution here, if you've got one.

pathawks:
Oh! I understand.

Odd that it breaks validation; I've never run into that before. May I ask where you saw this? HTML-Proofer doesn't seem to mind.

benbalter:
> May I ask where you saw this? HTML-Proofer doesn't seem to mind.

Looking at https://github.com/benbalter/jekyll-seo-tag/issues/4, I want to say it was [the w3 validator service](https://validator.w3.org/)?

pathawks:
I am trying to get to the bottom of this, but am having trouble reproducing the warning.

The [W3C's Nu Html Checker](https://validator.w3.org/nu/) reports no errors or warnings with the following HTML document.

```html
<!DOCTYPE html>
<html><head><title>Test</title></head>
<body>
<h1>“Smart Quotes”</h1>
<meta property="test" content="“Smart Quotes”" />
</body>
</html>
```

benbalter:
Huh. I'm scratching my head now. Maybe it was a HTML proofer bug? Do tests pass without the Kramdown option passed? I'm all for simplicity, but I'd rather make the code slightly more complex, if that means not making things more complex for the user (e.g., does not produce valid HTML out of the box).

pathawks:
HTML-Proofer [has no complaints](https://travis-ci.org/benbalter/jekyll-seo-tag/builds/101694480#L242-L243)

benbalter:
Huh. Odd. Could have sworn I had trouble before, but who knows. Can't reproduce it now. Thanks for talking through the change with me.

PR #22, opened 2016-01-10
Only process Liquid template once

pathawks:
Very small performance optimization

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -11,7 +11,7 @@ module Jekyll
 
     def render(context)
       @context = context
-      output = Liquid::Template.parse(template_contents).render!(payload, info)
+      output = template.render!(payload, info)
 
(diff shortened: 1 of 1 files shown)


lib/jekyll-seo-tag.rb
```diff
@@ -35,8 +35,8 @@ def info
       }
     end
 
-    def template_contents
-      @template_contents ||= File.read(template_path).gsub(/(>\n|[%}]})\s+(<|{[{%])/,'\1\2').chomp
+    def template
+      @template ||= Liquid::Template.parse(File.read(template_path).gsub(/(>\n|[%}]})\s+(<|{[{%])/,'\1\2').chomp)
```
parkr: Maybe keep `template_contents` to encapsulate the reading, and add `template` to encapsulate the parsing :)

pathawks: Good call

benbalter:
Nice. Thanks @pathawks!

Issue #23, opened 2016-01-11
jekyll-feed


akhyarrh:
I don't know how to give better title.

### `jekyll-feed` use `site.name:`, not `site.title:`
```yaml
# Should I use this ?
# Look funny ;)
title:     'Jekyll'
name: 'Jekyll'
```

### Author
```yaml
# post front-matter

# jekyll-seo-tag
author: John Doe

# jekyll-feed
author:
  name: John Doe
  uri: john-doe.com
```

Will there be a conflict?

pathawks:
We should use `site.title` if it exists, and fallback to `site.name` otherwise. This would mirror behavior of jekyll/jekyll-feed#72

If `author` is a hash, we should use `author.name` instead. This would mirror behavior of jekyll/jekyll-feed#31

benbalter:
Fixed via https://github.com/benbalter/jekyll-seo-tag/issues/25. FWIW, I suspect they should all be `site.title` to page `post.title` and `page.title`, but makes sense to have the fallback to avoid a breaking change.

PR #25, opened 2016-01-12
Use site.name if site.title is not present

pathawks:
This makes our behavior match that of **jekyll/jekyll-feed**

#23


benbalter:
Nice work @pathawks. :baby_chick: :vhs: :ice_cream:

Issue #18, opened 2016-01-06
Author name == Twitter Username?


pathawks:
We are currently assuming that, if an `author` is provided for a post, the `author` is [a string that corresponds to a Twitter username.](https://github.com/benbalter/jekyll-seo-tag/blob/v0.1.3/lib/template.html#L92-L94)

That is to say, `author: "Pat Hawks"` would cause false data to be rendered, because my Twitter username has no space in it, but I *do* include a space in the name I include on my site.

Not sure of the best path forward. We could assume `author` is a hash that has keys `twitter` and `name`. We could look up the provided `author` name in a hash in `_data/`, or something...

benbalter:
It adds some complexity, but I'd be fine taking [the approach Jekyll-Feed takes](https://github.com/jekyll/jekyll-feed/blob/98cd38e7220b7e692f7f0f1186228a5673267adf/README.md#optional-configuration-options), meaning, it can either be a string (the user's username) or a hash, with some sort of username property, both falling back to the global config, if none is present.

benbalter:
Fixed by #26 / #52

PR #26, opened 2016-01-12
Allow author data as an object

pathawks:
Try to mirror behavior of **jekyll-feed**

#18 #23

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -44,6 +44,23 @@
   {% assign seo_description = seo_description | markdownify | strip_html | strip_newlines | escape_once %}
 {% endif %}
 
+{% if page.author %}
+  {% if page.author.name %}
+    {% assign seo_author_name = page.author.name %}
+  {% else %}
+    {% assign seo_author_name = page.author %}
+  {% endif %}
+
+  {% if page.author.twitter %}
+    {% assign seo_author_twitter = page.author.twitter %}
+  {% else %}
+    {% assign seo_author_twitter = page.author %}
+  {% endif %}
+{% endif %}
+{% if seo_author_twitter %}
+  {% assign seo_author_twitter = seo_author_twitter | replace:"@","" | prepend:"@" %}
+{% endif %}
+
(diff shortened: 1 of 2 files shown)


lib/template.html
```diff
@@ -35,6 +35,20 @@
...
+  {% if page.author.name %}
+    {% assign seo_author_name = page.author.name %}
+  {% else %}
+    {% assign seo_author_name = page.author %}
+  {% endif %}
+
+  {% if page.author.twitter %}
+    {% assign seo_author_twitter = page.author.twitter | replace:"@","" | prepend:"@" %}
+  {% else %}
+    {% assign seo_author_twitter = page.author | replace:"@","" | prepend:"@" %}
```
benbalter: 1. It looks like you're adding the `@`? I believe the Twitter spec says it shouldn't have an `@`, but I haven't looked at it in a while. 

2. We can DRY things up a bit by moving the normalization down to the one place it's used.

pathawks: > It looks like you're adding the `@`?

[It was there before.](https://github.com/benbalter/jekyll-seo-tag/pull/26/files#diff-b18f6b02baa17fff6667f45aca6143c5L101)

The [example at the bottom of this page](https://dev.twitter.com/cards/getting-started) includes the `@`

pathawks:
> We can DRY things up a bit by moving the normalization down to the one place it's used.

All of our other variables are normalized before they are used, and I wanted to keep with that convention.

I could do something like this:

```html
{% if page.author.twitter %}
  {% assign seo_author_twitter %}
{% else %}
  {% assign seo_author_twitter %}
{% endif %}
{% if seo_author_twitter %}
  {% assign seo_author_twitter = page.author.twitter | replace:"@","" | prepend:"@" %}
{% endif %}
```

benbalter:
Nice. Sorry for the delay. Also opened https://github.com/benbalter/jekyll-seo-tag/issues/31 to piggy back on this, so that author's metadata can be centralized, rather than in each post.

PR #28, opened 2016-01-12
Consolidate and clean up JSON-LD

pathawks:
- [x] Move all JSON-LD tests into one section
- [x] Consolidate all JSON-LD `<script>` tags into just one block
- [x] Modify the way `@type` is decided for a page
  - [x] Decide on a default `@type` (Maybe different for home vs. post pages?)
  - [x] Provide a sensible way to override the default `@type`
  - [x] Document all changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,6 @@ The SEO tag will respect any of the following if included in your site's `_confi
 
 * `logo` - Relative URL to a site-wide logo (e.g., `assets/your-company-logo.png`)
 * `social` - For [specifying social profiles](https://developers.google.com/structured-data/customize/social-profiles). The following properties are available:
-  * `type` - Either `person` or `organization` (defaults to `person`)
   * `name` - If the user or organization name differs from the site's name
   * `links` - An array of links to social media profiles.
 * `google_site_verification` for verifying ownership via Google webmaster tools
@@ -145,3 +144,12 @@ There are several ways to convey this author-specific information. Author inform
   ```yml
   author: benbalter
   ```
+
+### Customizing JSON-LD output
+
+The following options can be set for any particular page. While the default options are meant to serve most users in the most common circumstances, there may be situations where more precise control is necessary.
+
+* `seo`
+  * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
+  * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
+  * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
--- a/lib/template.html
+++ b/lib/template.html
@@ -1,5 +1,9 @@
 <!-- Begin Jekyll SEO tag v{{ seo_tag.version }} -->
 
+{% if page.url == "/" or page.url == "/about/" %}
+  {% assign seo_homepage_or_about = true %}
+{% endif %}
+
 {% if site.url %}
   {% assign seo_url = site.url | append: site.baseurl %}
 {% endif %}
@@ -20,6 +24,17 @@
   {% if site.description %}
     {% assign seo_title = seo_title | append:" - " | append: site.description %}
   {% endif %}
+{% endif %}
+
+{% if page.seo and page.seo.name %}
+    {% assign seo_name = page.seo.name %}
+{% elsif seo_homepage_or_about and site.social and site.social.name %}
+    {% assign seo_name = site.social.name %}
+{% elsif seo_homepage_or_about and seo_site_title %}
+    {% assign seo_name = seo_site_title %}
+{% endif %}
+{% if seo_name %}
+  {% assign seo_name = seo_name | markdownify | strip_html | strip_newlines | escape_once %}
 {% endif %}
 
 {% if seo_title %}
@@ -51,6 +66,22 @@
     {% endif %}
   {% endif %}
   {% assign seo_author_twitter = seo_author_twitter | replace:"@","" %}
+{% endif %}
+
+{% if page.seo and page.seo.type %}
+    {% assign seo_type = page.seo.type %}
+{% elsif seo_homepage_or_about %}
+    {% assign seo_type = "WebSite" %}
+{% elsif page.date %}
+    {% assign seo_type = "BlogPosting" %}
+{% else %}
+    {% assign seo_type = "WebPage" %}
+{% endif %}
+
+{% if page.seo and page.seo.links %}
+    {% assign seo_links = page.seo.links %}
+{% elsif seo_homepage_or_about and site.social and site.social.links %}
+    {% assign seo_links = site.social.links %}
 {% endif %}
 
 {% if seo_tag.title and seo_title %}
@@ -125,22 +156,12 @@
   {
     "@context" : "http://schema.org",
 
-{% if page.seo and page.seo.type %}
-    "@type" : {{ page.seo.type | jsonify }},
-{% elsif page.url == "/" %}
-    "@type" : "WebSite",
-{% elsif page.date %}
-    "@type" : "BlogPosting",
-{% else %}
-    "@type" : "WebPage",
+{% if seo_type %}
+    "@type" : {{ seo_type | jsonify }},
 {% endif %}
 
-{% if page.seo and page.seo.name %}
-    "name" : {{ page.seo.name | jsonify }},
-{% elsif page.url == "/" and site.social and site.social.name %}
-    "name" : {{ site.social.name | jsonify }},
-{% elsif page.url == "/" and seo_site_title %}
-    "name" : {{ seo_site_title | jsonify }},
+{% if seo_name %}
+    "name" : {{ seo_name | jsonify }},
 {% endif %}
 
 {% if seo_page_title %}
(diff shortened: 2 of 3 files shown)


pathawks:
Previously, `@type` could be set to one of [`WebSite`](https://github.com/benbalter/jekyll-seo-tag/blob/v0.1.4/lib/template.html#L61), [`NewsArticle`](https://github.com/benbalter/jekyll-seo-tag/blob/v0.1.4/lib/template.html#L83), [`Organization`](https://github.com/benbalter/jekyll-seo-tag/blob/v0.1.4/lib/template.html#L109), or [`person` or **`{{site.social.type}}`**](https://github.com/benbalter/jekyll-seo-tag/blob/v0.1.4/lib/template.html#L120).

I believe we are only allowed one `@type` per page.

It seems that this should probably be different depending on what the page *is.* 

For a site's homepage, `WebSite` would be an obvious default. `Person` might make sense for a personal blog, or `Organization` for an organization's website.

For post pages, `BlogPosting` or `NewsArticle` could make sense, but how to determine which one?

For other pages, maybe `WebPage`?

I would like to eliminate any *magic*. Perhaps it would be best to default to the most general type for each page (`WebSite` or `WebPage`), and provide an easy way to override the type for a given page (`{{ page.seo.type }}`?)

Thoughts?

benbalter:
> I would like to eliminate any magic. 

:sparkling_heart:. Completely agree.

I think posts can be `BlogPosting` (if that's what Google supports, for some reason I thought they only supported newsArticle), by default, with an easy override.

For the homepage, again, webSite, with Person or Organization being overridden by the config.

For pages, webPage makes sense.

What do you think about collections? I'm thinking nothing unless specified?

Really glad you're digging into many of the (erroneous) assumptions I made. If you're a GitHub Pages user, you'll notice the plugin has been silently made available to all GitHub Pages users.

pathawks:
> if that's what Google supports

[Google supports](https://developers.google.com/structured-data/rich-snippets/articles#article_markup_properties) both `NewsArticle` and `BlogPosting`.

> What do you think about collections?

It seems that a page in a collection is still a WebPage?

> you'll notice the plugin has been silently made available to all GitHub Pages users.

That's how I discovered the plugin: trolling the **pages-gem** changelog :wink:

lib/template.html
```diff
@@ -111,27 +93,56 @@
...
 
+<script type="application/ld+json">
+  {
+    "@context" : "http://schema.org",
+
+{% if page.seo and page.seo.type %}
+    "@type" : {{ page.seo.type | jsonify }},
+{% elsif page.url == "/" %}
+    "@type" : "WebSite",
+{% elsif page.id %}
```
pathawks: There *must* be a better way to check if the current `page` is a **Post** :confounded:

lib/template.html
```diff
@@ -111,27 +93,56 @@
...
+    "logo": {{ site.logo | prepend: "/" | prepend: seo_url | jsonify }},
 {% endif %}
 
+{% if page.seo and page.seo.links %}
+    "sameAs" : {{ page.seo.links | jsonify }},
+{% elsif page.url == "/" and site.social and site.social.links %}
+    "sameAs" : {{ site.social.links | jsonify }},
+{% endif %}
+
+{% if page.url == "/" and site.social and site.social.name %}
```
pathawks: This feels like :tophat:Magic. I wasn't sure if `site.social.name` is an existing thing that we need to continue to support, if If I could just check for something like `site.seo.name` instead.

lib/template.html
```diff
@@ -111,27 +93,56 @@
...
-      "url" : {{ seo_url | jsonify }},
-      "sameAs" : {{ site.social.links | jsonify }}
-    }
-  </script>
+    "logo": {{ site.logo | prepend: "/" | prepend: seo_url | jsonify }},
 {% endif %}
 
+{% if page.seo and page.seo.links %}
+    "sameAs" : {{ page.seo.links | jsonify }},
+{% elsif page.url == "/" and site.social and site.social.links %}
```
pathawks: This feels like :tophat:Magic. I wasn't sure if `site.social.links` is an existing thing that we need to continue to support as a fallback.

lib/template.html
```diff
@@ -111,27 +93,56 @@
...
+    "sameAs" : {{ site.social.links | jsonify }},
+{% endif %}
+
+{% if page.url == "/" and site.social and site.social.name %}
+    "name" : {{ site.social.name | jsonify }},
```
pathawks: Looks like we're trying to output `name` [twice](https://github.com/benbalter/jekyll-seo-tag/pull/28/files#diff-b18f6b02baa17fff6667f45aca6143c5R111). Whoops. :rage1:

lib/template.html
benbalter: I'd check if it has a date? Even if it's a collection with a date (possible in 3.0), it'd be closer to a post (dated, chronological), than a page (unordered, undated).

benbalter:
@pathawks still interested in seeing this happen?

pathawks:
> still interested in seeing this happen?

I am. Working on rebasing. The tests have changed quite a bit.

spec/jekyll_seo_tag_spec.rb
```diff
@@ -124,7 +124,6 @@
 
       it 'outputs the logo' do
         expect(json_data['logo']).to eql('http://example.invalid/logo.png')
-        expect(json_data['url']).to eql('http://example.invalid')
```
pathawks: `url` [should be the URL of the current page, not the site](https://schema.org/BlogPosting). Since it doesn't have anything to do with outputting the logo anyway, I just removed this check.

pathawks:
@benbalter Ok, how does everything look?

spec/jekyll_seo_tag_spec.rb
benbalter: :+1:

README.md
```diff
@@ -81,6 +80,10 @@ The SEO tag will respect the following YAML front matter if included in a post,
 * `description` - A short description of the page's content
 * `image` - Relative URL to an image associated with the post, page, or document (e.g., `assets/page-pic.jpg`)
 * `author` - Page-, post-, or document-specific author information (see below)
+* `seo` - The following properties are available:
+  * `type` - What this particular page represents. Can any type supported by [Schema.org](http://schema.org/docs/schemas.html) (i.e.: `WebSite`, `BlogPosting`, `NewsArticle`, `Person`, `Organization`, etc.)
+  * `name` - The name of the post, page, or document. This is similar to `title` (which is used as the default value), but will differ in some cases, i.e. the page represents an organization or a person.
```
benbalter: When would/should the JSON-LD name differ from page title?

benbalter:
@pathawks this is awesome, and thanks for working through the merge conflicts in the tests. :grimacing: 

Three suggestion:

1. I'd imagine for the majority of use cases, the site as a whole would represent either a user or an organization (if you use the term organization broadly), and if social type/links are present, we should publish them on the home page. The alternative, placing that in the front matter of index page feels strange to me, no? The other use case would be an `/about` page. Should that also pull from the site-wide config, if no individual config is present (the alternative likely being to have the same info twice).

2. Both site-wide, and in a per-page basis, when would the page/site title differ from the JSON-LD title we'd want? Struggling to think through the sufficient use case that would justify the option.

3. What would you think about moving the JSON-LD docs to a separate "customizing JSON-LD output" or similar section? The thinking is, for most people, they shouldn't have to do anything. If the metadata is there, and e.g., it's a post, it should just work. Type, links, etc. are available, but I suspect they're more power user features (as it requires a least a basic understanding of the value of linked data, that I don't know we should assume of every user)?

README.md
pathawks: **Title:** Welcome to Joe's Café
**Name:** Joe's Café

benbalter: Ugh. Yeah, that makes sense. Thanks for walking me through it.

pathawks:
> Both site-wide, and in a per-page basis, when would the page/site title differ from the JSON-LD title we'd want?

"About Frank's Burgers" vs. "Frank's Burgers"

> What would you think about moving the JSON-LD docs to a separate "customizing JSON-LD output" or similar section?

Yes, this is definitely power user territory. I will work on that :+1:

benbalter:
> "About Frank's Burgers" vs. "Frank's Burgers"

:100:% with you on that. Thanks for explaining.

pathawks:
Had to rebase everything; it was getting to be too much. :rage3:
Kept the PR from @CGarces

The old history can still be found [here](https://github.com/benbalter/jekyll-seo-tag/compare/master...pathawks:3d24727b6910469cfd392d586ed1692c6bab5a92)

pathawks:
> I'd imagine for the majority of use cases, the site as a whole would represent either a user or an organization
...

What if I have a multi-author blog, and each author has her own bio page? Jane's bio should be able to indicate that this is the same Jane as this user on GitHub, etc.

I agree that this is a weird edge case, but I still feel like there should be an option to override via page front matter.

benbalter:
> I agree that this is a weird edge case, but I still feel like there should be an option to override via page front matter.

Totally with you, and `about/ben` and `about/pat` is exactly what I had in mind. I meant for your example, if we had JekyllExperts.com, and want to store our @JekyllExperts twitter handle and LinkedIn profile, it feels more natural to store it in the config, rather than in `/index.md` and `/about.md` twice. Is it too magic to auto include the `config.yml` version in `/` and `/about`?

pathawks:
> Is it too magic to auto include the config.yml version in / and /about?

OH! I think I understand what you are saying.

```diff
  {% if page.seo and page.seo.links %}
      "sameAs" : {{ page.seo.links | jsonify }},
- {% elsif page.url == "/" and site.social and site.social.links %}
+ {% elsif (page.url == "/" or page.url == "about") and site.social and site.social.links %}
      "sameAs" : {{ site.social.links | jsonify }},
  {% endif %}
```

:grey_question:

benbalter:
> OH! I think I understand what you are saying.

Yeah, does that make sense? (it may not)

pathawks:
> Yeah, does that make sense?

Yessir. I will experiment later :+1:

pathawks:
`sameAs` Links will now also appear on `/about/` page :+1:

benbalter:
Nice work, once again @pathawks! Thanks. :link: :cake: :cactus:

PR #29, opened 2016-01-13
Make <title> optional

pathawks:
I am *very* picky about my `<title>` tags.

This PR provides a way to disable the plugin from outputting the `<title>` tag so that I can continue to use my own.

```html
{% seo title:false %}
```

By putting the option inside the tag, we avoid :christmas_tree:

Does this seem reasonable?

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -3,6 +3,13 @@ module Jekyll
 
     attr_accessor :context
 
+    def initialize(_, markup, _)
+      super
+      @options = {
+        "title" => !(markup =~ /title\s*:\s*false/i)
+      }
+    end
+
     def render(context)
       @context = context
       output = template.render!(payload, info)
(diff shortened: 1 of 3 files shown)


benbalter:
@pathawks what would you want the title to be, if not the default? Do you need to be able to disable it on a per-page basis, or site wide?

pathawks:
> what would you want the title to be, if not the default?

On my personal blog, the page title is simply `{{ page.title }}` with nothing else tacked on. My site title is just my name, and it felt weird having my name plastered all over *everything*.

Further more, I have to customize the title to work with [jekyll/jekyll-archives](https://github.com/jekyll/jekyll-archives) since, last I checked, it does not provide a default title. That, however, is something that *could* be within the scope of this plugin to work around.

> Do you need to be able to disable it on a per-page basis, or site wide?

Because of the way my template is setup, it would probably be all or nothing. I have a `<head>` include file that is included in all pages. In that include, I would have my `<title>` and the `{% seo %}` tag.

Issue #31, opened 2016-01-20
Support for global author's data in _data


benbalter:
Where the object added in https://github.com/benbalter/jekyll-seo-tag/pull/26, can be replaced with a name, which would pull `site.data.authors[name]`.

pathawks:
**YES PLEASE** :+1:

parkr:
:+1: on this too!

pathawks:
I can take a stab at this over the weekend.

ickc:
:+1:

benbalter:
Fixed by #52

PR #33, opened 2016-01-23
Lookup author in site.data.authors

pathawks:
Where will the plugin try to find a post author's Twitter username?

1. `site.data.authors[post.author].twitter`
2. `site.data.authors[post.author].name`
3. `post.author.twitter`
3. `post.author.name`
3. `post.author`
1. `site.data.authors[site.author].twitter`
2. `site.data.authors[site.author].name`
3. `site.author.twitter`
3. `site.author.name`
3. `site.author`

*Ugh.* I don't even begin to know how to document this. It is too bad that there isn't just one standard convention for all of this.

#31

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -36,11 +36,36 @@
 {% endif %}
 
 {% if page.author %}
-  {% assign seo_author_name = page.author.name | default: page.author %}
-  {% assign seo_author_twitter = page.author.twitter | default: page.author %}
+  {% if site.data.authors[page.author] %}
+    {% assign seo_author_name = site.data.authors[page.author].name %}
+    {% assign seo_author_twitter = site.data.authors[page.author].twitter | default: seo_author_name %}
+  {% elsif page.author.name %}
+    {% assign seo_author_name = page.author.name %}
+    {% assign seo_author_twitter = page.author.twitter | default: seo_author_name %}
+  {% elsif page.author.twitter %}
+    {% assign seo_author_twitter = page.author.twitter %}
+    {% assign seo_author_twitter = page.author.twitter %}
+  {% else %}
+    {% assign seo_author_name = page.author %}
+    {% assign seo_author_twitter = page.author %}
+  {% endif %}
+{% elsif site.author %}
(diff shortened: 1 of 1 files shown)


benbalter:
Looking through things, do we ever actually use `seo_author_name`? It looks like the only author information that's used is for the Twitter Card, which requires a twitter username, not a human-readable name. That should let us cut things in half.

We may be able to simplify things a bit further:

1. Set `seo_author` to `page.author` or if that doesn't exist `site.author`
2. If `seo_author.twitter` exists (it's an object), use that 
3. If `site.authors[seo_author].twitter` exists, use that
4. `seo_author`

Does that account for every possible permutation?

pathawks:
> Looking through things, do we ever actually use `seo_author_name`?

Not currently, but I would like to.

Honestly, it kind of feels like *magic* to assume that the author name will also just happen to match that user's Twitter username. Here is how I would really like to do things (in Ruby/pseudocode):

```ruby
author = {}

if post.author.is_a? Hash
  author = post.author

elsif site.data.authors[post.author].is_a? Hash
  author = site.data.authors[post.author]

elsif post.author.is_a? String
  author = { 'name' : post.author }

elsif site.author.is_a? Hash
  author = site.data.authors[post.author]

elsif site.author.is_a? String
  author = { 'name' : site.author.to_s }

end
```

Then, we can use `author.name` or `author.twitter` or `author.url` or whatever anywhere we want.

As far as I know, Liquid has no way to check the type of an object or create hashes, so I'm not really sure of the best way to go about this...

benbalter:
Yeah, looks like it's going to be a lot more sane in Ruby, and it's just string manipulation, so I don't think there's much security concern.

How about something like:

```
author = post.author || site.author # Default to post.author, fallback to site.author, both stings and objects
return author.twitter if author.twitter # The author is an object, use the Twitter property
return site.data.authors[author].twitter if site.data.authors[author].twitter # The author is a string, check for a matching key in the data 
author # The author is a string with no matching data, this is all we've got
```

That said, where would we use author.name within the plugin?

PR #36, opened 2016-01-29
Add `default` filter

pathawks:
It occurred to me that using the `default` filter in Liquid could allow us to dramatically simplify the logic of something like #33 

Unfortunately, the `default` filter is not available in Jekyll 2, so I have provided it for compatibility.

I am sure that I am not namespacing properly. `Filters` needs to be a module somewhere, I am just not exactly sure where to put it.


parkr:
:+1:

benbalter:
Nice. So much simpler. Thanks again, @pathawks!

PR #39, opened 2016-01-31
Remove ugly blank lines at header

CGarces:


pathawks:
Aren’t these lines [already filtered](https://github.com/benbalter/jekyll-seo-tag/blob/master/lib/jekyll-seo-tag.rb#L34) from the output?

CGarces:
Opps, sorry.

I'm using the [template directly](https://github.com/CGarces/CGarces.github.io/blob/master/_includes/seo_template.html) on my site. I not use the ruby code.

Sorry for the mistake.

pathawks:
Thanks :+1:

PR #40, opened 2016-02-01
Use seo_description instead of page.description

CGarces:
seo_description is used in  the meta tags but not in the ld+json
If page.description is not set, the the description at ld+json will be null.


benbalter:
Makes sense. Thanks @CGarces!

PR #41, opened 2016-02-01
Use seo_description instead of page.description

CGarces:
seo_description is used in the meta tags but not in the ld+json
If page.description is not set, the the description at ld+json will be null.

Meta tags and ld+json should use the same logic.
Same PR (#40) was send to main branch.


benbalter:
Thanks @CGarces! Believe the failing tests are due to https://github.com/benbalter/jekyll-seo-tag/pull/43.

PR #43, opened 2016-02-03
Update tests to work with Liquid 3

### LINKED ISSUE #42: Tests fail
I was starting to make a pull request for #24, but after forking and cloning `master` all tests fail with:

```
NoMethodError:
       private method `new' called for Jekyll::SeoTag:Class
```

I tried several things (ruby versions, gem re-installs) and then tested the 0.1.4 tag with the same test failures. Just to be sure I pushed a small readme change to travis and the build failed on travis, too: https://travis-ci.org/boturnbow/jekyll-seo-tag/builds/106609578. I cloned jekyll/jekyll and benbalter/jekyll-avatar and ran tests locally - all passed. So, that leads me to believe that the problem is not entirely a local-to-my-machine problem.

Any insights?

pathawks:
Fixes #42

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -26,8 +26,10 @@ def page(options={})
 end
 
 def post(options={})
-  page = Jekyll::Post.new site, CONFIG_DEFAULTS["source"], "", "2015-01-01-post.md"
-  page.data = options
+  filename = File.expand_path("2015-01-01-post.md", CONFIG_DEFAULTS["source"])
+  config = { :site => site, :collection => site.collections["posts"] }
+  page = Jekyll::Document.new filename, config
+  page.merge_data!(options)
   page
 end
 
--- a/spec/jekyll_seo_tag_spec.rb
+++ b/spec/jekyll_seo_tag_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe Jekyll::SeoTag do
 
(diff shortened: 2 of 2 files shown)


pathawks:
`Jekyll::Post` Does not exist in **Jekyll 3**

spec/spec_helper.rb
```diff
@@ -26,8 +26,10 @@ def page(options={})
 end
 
 def post(options={})
-  page = Jekyll::Post.new site, CONFIG_DEFAULTS["source"], "", "2015-01-01-post.md"
-  page.data = options
+  site = site()
+  config = { :site => site, :collection => site.collections["posts"] }
```
pathawks: Does this make sense? I didn't want to create two `site` objects.

spec/spec_helper.rb
```diff
@@ -26,8 +26,11 @@ def page(options={})
 end
 
 def post(options={})
-  page = Jekyll::Post.new site, CONFIG_DEFAULTS["source"], "", "2015-01-01-post.md"
-  page.data = options
+  site = site()
```
pathawks: I need to use `site` twice in `config` and wanted to avoid creating two objects. Does this make sense? Is this ok?

pathawks: Doesn't even matter, I suppose. `context` creates two `site` objects indirectly.

benbalter:
Ahh. Nice work. Thanks @pathawks!

PR #48, opened 2016-02-03
Add site values as defaults for author

CGarces:
Use site options as default if not is set by post.


benbalter:
@CGarces thanks for the pull request! I implemented this as part of #52. :smile:

PR #52, opened 2016-02-20
Support for global author data

benbalter:
This PR builds on @pathawks work over in https://github.com/benbalter/jekyll-seo-tag/pull/33, by moving the author logic to Ruby where things can be implemented a bit more cleanly.

Still todo:

- [x] tests
- [x] docs

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,10 @@
+Metrics/LineLength:
+  Exclude:
+    - spec/**/*
+    - jekyll-seo-tag.gemspec
+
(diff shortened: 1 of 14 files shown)


benbalter:
@pathawks as a third pass, I actually moved things back to Liquid, but tests still pass. Documented things in the README. Does this solve all the use cases you set out to solve in #33?

benbalter:
Not sure that this is the right implementation, but hoping to give this a spin in the wild to get a better sense for how it feels in practice.

PR #54, opened 2016-02-21
Make <title> Optional

benbalter:
Based on @pathawks's https://github.com/benbalter/jekyll-seo-tag/pull/29. With the following additions:

* Merged in `master`, which changed the way tests are implemented
* Use the documented `initialize` argument names for consistency
* Move `options` and `title?` to their own methods for simplicity
* Use the (now) existing `seo_tag`, rather than the `seo` context namespace
* Use `title=false` rather than `title: false` to retain consistency with things like the include tag, which uses the `key=value` format for passing argument to tags.
* Add documentation


pathawks:
:heart_eyes: **Thank you!**

PR #56, opened 2016-02-22
Support article:publisher without fb:app_id

pathawks:
Docs say these can be used independently. This PR allows one to be set without outputting an empty tag for the other.


benbalter:
Makes sense. Thanks @pathawks!

PR #44, opened 2016-02-03
Make page.image url relative to site rather than absolute

boturnbow:
Fixes #24 

Tests pass when https://github.com/benbalter/jekyll-seo-tag/pull/43 is merged,

PR #57, opened 2016-02-23
Update twitter:image to output full url

boturnbow:
Oops - I should have updated the twitter image in https://github.com/jekyll/jekyll-seo-tag/pull/44 Now it matches behavior of og:image.

### DIFF
--- a/spec/jekyll_seo_tag_spec.rb
+++ b/spec/jekyll_seo_tag_spec.rb
@@ -247,6 +247,16 @@ describe Jekyll::SeoTag do
           expect(output).to match(expected)
         end
       end
+
+      context 'with page.image' do
+        let(:site) { make_site('twitter' => site_twitter, 'url' => 'http://example.invalid') }
+        let(:page) { make_page('image' => 'foo.png') }
+
+        it 'outputs the image' do
+          expected = %r{<meta name="twitter:image" content="http://example.invalid/foo.png" />}
+          expect(output).to match(expected)
+        end
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/jekyll_seo_tag_spec.rb
```diff
@@ -247,6 +247,16 @@
           expect(output).to match(expected)
         end
       end
+
+      context 'with page.image' do
+        let(:site) { make_site('twitter' => site_twitter, 'url' => 'http://example.invalid') }
+        let(:page) { make_page('image' => 'foo.png') }
```
benbalter: Should the image be relative to the root or to the page? I suspect you'd rarely want an image to be at `/2016/01/01/foo/foo.png`?

boturnbow: Root. I think most folks would place images in /img/ or /assets/. But the config or code does not preclude either use. Would it be clearer if that kind of usage were in the tests?

benbalter: If root, then you don't need `prepend: '/'` above. I'd have the test match your intended use, as right now, if I put `/img/foo.png`, it'd output as `http://exmaple.invalid//img/foo.png`, I believe.

boturnbow: Yes, it would. In my change for page image I copied the usage for site logo which prepends a / (https://github.com/jekyll/jekyll-seo-tag/blob/master/lib/template.html#L149) and documents it's use in that manner. I, personally, prefer starting the config value with a / and would be happy to make that change to both page image and site logo.

benbalter: >  I, personally, prefer starting the config value with a / and would be happy to make that change to both page image and site logo.

:+1:

boturnbow:
Pull request now contains a breaking change for site logo and page image - both are expected to be configured starting with a /.

benbalter:
Looks good. Thanks @boturnbow! Given how young the plugin is, I'm fine with a breaking change in machine-readable metadata to get a sane default going forward.

PR #60, opened 2016-02-26
Update JSON-LD output to current usage for site logo and page image

boturnbow:


lib/template.html
```diff
@@ -181,7 +181,7 @@
 {% endif %}
 
 {% if site.logo %}
-    "logo": {{ site.logo | prepend: "/" | prepend: seo_url | jsonify }},
+    "logo": {{ site.logo | prepend: seo_url | jsonify }},
```
pathawks: How many times do we use `site.logo`? Would it make sense to `assign` it at the top of the template?

boturnbow: This is the only time that logo is used. Page image would be a better candidate for a variable as it is used 3 times with the same filters each time.

boturnbow:
@pathawks are you interested in a change where both page.image and site.logo are set as vars at the top of the template? i.e. seo_page_image, seo_site_logo

pathawks:
> @pathawks are you interested in a change where both page.image and site.logo are set as vars at the top of the template?

Yes, this makes me happy. :+1: :shipit:

benbalter:
Nice work. Thanks @boturnbow!

Issue #64, opened 2016-03-01
Issues with the plugin


XhmikosR:
So, I'm trying to get the plugin in Bootstrap v4-dev branch and I notice the following (sorry for not splitting them)

1. `<meta name="twitter:description" content="" />` while `<meta name="description" content="Bootstrap, a sleek, intuitive, and powerful mobile first front-end framework for faster and easier web development.">`. The description is set in _config.yml
2. The title is wrong if I set `title=false`. `<meta property="og:title" content="Introduction" />` and `"headline": "Introduction",` vs `<title>Introduction &middot; Bootstrap</title>`. 

The branch for testing is here: https://github.com/twbs/bootstrap/compare/v4-dev...v4-dev-xmr-jekyll-social-tag

pathawks:
> The title is wrong if I set title=false. <meta property="og:title" content="Introduction" /> and "headline": "Introduction", vs <title>Introduction &middot; Bootstrap</title>.

But the title of the page _is_ “Introduction”

XhmikosR:
The title is `Introduction &middot; Bootstrap`, not only `Introduction`.

pathawks:
Yes, I understand that that is what is contained in the `<title>` tag, but that isn’t _really_ the title of the page. If you were to list all of the pages in a table of contents, they wouldn’t all have “Bootstrap” tacked onto the end.

Twitter limits this field to 70 characters, so it is important to avoid cluttering this with anything extra.

XhmikosR:
I see. Then that makes sense.

The first issue is valid though.

XhmikosR:
Although, I do have a case where I see both types.

```html
<title>Downloads &middot; MPC-HC</title>
<meta name="description" content="Download our binaries, source code, etc.">

        <!-- Begin Jekyll SEO tag v1.3.1 -->
<meta property="og:title" content="Downloads" />
<meta name="description" content="Download our binaries, source code, etc." />
<meta property='og:description' content="Download our binaries, source code, etc." />
<link rel="canonical" href="https://mpc-hc.org/downloads/" />
<meta property='og:url' content='https://mpc-hc.org/downloads/' />
<meta property="og:site_name" content="MPC-HC" />
<meta name="twitter:card" content="summary" />
<meta name="twitter:site" content="@mpc_hc" />
<meta name="twitter:title" content="Downloads - MPC-HC" />
<meta name="twitter:description" content="Download our binaries, source code, etc." />
[...]

PR #65, opened 2016-03-02
Fix og:title condition and remove redundant tags

pathawks:
In investigating #64, I learned that:

 1. We were not checking for `seo_page_title` properly
 2. We [do not need to repeat Open Graph tags just for Twitter](https://dev.twitter.com/cards/getting-started#opengraph)

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -126,12 +126,6 @@
 {% if site.twitter %}
   <meta name="twitter:card" content="summary" />
   <meta name="twitter:site" content="@{{ site.twitter.username | replace:"@","" }}" />
-  <meta name="twitter:title" content="{{ seo_title }}" />
-  <meta name="twitter:description" content="{{ seo_description }}" />
-
-  {% if page.image %}
-    <meta name="twitter:image" content="{{ page.image | prepend: seo_url | escape }}" />
-  {% endif %}
 
   {% if seo_author_twitter %}
     <meta name="twitter:creator" content="@{{ seo_author_twitter }}" />
--- a/lib/template.html
+++ b/lib/template.html
@@ -88,7 +88,7 @@
   <title>{{ seo_title }}</title>
 {% endif %}
 
-{% if seo_title %}
+{% if seo_page_title %}
   <meta property="og:title" content="{{ seo_page_title }}" />
 {% endif %}
 
--- a/spec/jekyll_seo_tag_spec.rb
+++ b/spec/jekyll_seo_tag_spec.rb
@@ -266,16 +266,6 @@ EOS
           expect(output).to match(expected)
         end
       end
-
-      context 'with page.image' do
-        let(:site) { make_site('twitter' => site_twitter, 'url' => 'http://example.invalid') }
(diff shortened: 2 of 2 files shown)


lib/template.html
```diff
@@ -126,12 +126,6 @@
 {% if site.twitter %}
   <meta name="twitter:card" content="summary" />
   <meta name="twitter:site" content="@{{ site.twitter.username | replace:"@","" }}" />
-  <meta name="twitter:title" content="{{ seo_title }}" />
-  <meta name="twitter:description" content="{{ seo_description }}" />
```
pathawks: Twitter will use `og:title` and `og:description` instead :+1:

lib/template.html
```diff
@@ -126,12 +126,6 @@
...
-
-  {% if page.image %}
-    <meta name="twitter:image" content="{{ page.image | prepend: seo_url | escape }}" />
```
pathawks: Twitter will use `og:image` instead, no need to duplicate the same info here :+1:

XhmikosR:
My humble opinion, is that I'd rather be explicit in this case.

XhmikosR:
Imagine that Twitter changes its decision. This won't be backward-compatible.

pathawks:
> Imagine that Twitter changes its decision. This won't be backward-compatible.

Twitter could do a number of things that would make parsing websites more difficult for themselves. I don't know _why_ they would do this, and so it seems very unlikely.

It's not a matter of being explicit so much as being redundant.

XhmikosR:
It seems redundant, sure, but I'm not aware of any other SEO plugin that does this.

pathawks:
> If you’re already using Open Graph protocol to describe data on your page, it’s easy to generate a Twitter card without duplicating your tags and data. When the Twitter card processor looks for tags on your page, it first checks for the Twitter property, and if not present, falls back to the supported Open Graph property. This allows for both to be defined on the page independently, and minimizes the amount of duplicate markup required to describe your content and experience.

The reason that Twitter looks first for `twitter:description` is in case you'd like to specify a different description for Twitter rather than `og:description`. Since we are using the same description in both cases, there is no reason to include it twice. This according to Twitter's own documentation.

XhmikosR:
OK, I guess. I'm seeing this from my point of view where I don't yet use the plugin.

pathawks:
We’re working toward the same goal :beers:

One advantage of this being a plugin is that we can change our approach if Twitter should change its requirements in the future. :+1:

benbalter:
Wow. Learned so much from reading through this PR. Thanks for doing the digging @pathawks. Also learned why some use `name` and some use `property`. See https://xkcd.com/927/.

Based on the Twitter docs, which is the only thing we can go off of (as opposed to say, what they *might* do some point in the future), :+1: from me. Feel free to merge when you're happy with it. :smile:.

pathawks:
:white_check_mark:

Thank you both :smile_cat:

XhmikosR:
Can we have a patch release with this?

PR #70, opened 2016-03-03
Large image summary Twitter Card support

### LINKED ISSUE #68: Support Twitter summary card with large image
Hi,

Support of [large image in Twitter Summary Cards](https://dev.twitter.com/cards/types/summary-large-image) would be nice.

The only difference with the actual summary card seems to be the content attribute value:

```
<meta name="twitter:card" content="summary_large_image">
```
I manually added the meta:
https://github.com/sudweb/2016/blob/gh-pages/_includes/head.html#L8

I tested it on http://sudweb.github.io/2016/
It seems that if both twitter meta are present, Twitter use the large image version:

twitter-card-summary-large

I can try to submit a PR if you agree adding this.

Thanks.

DirtyF:
@benbalter here is a proposal to fix #68 but it fails the tests :/

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -127,7 +127,7 @@
 
   {% if page.image %}
     <meta name="twitter:card" content="summary_large_image" />
-  {% elsif  %}
+  {% else %}
     <meta name="twitter:card" content="summary" />
   {% endif %}
   
--- a/lib/template.html
+++ b/lib/template.html
@@ -124,7 +124,13 @@
 {% endif %}
 
 {% if site.twitter %}
-  <meta name="twitter:card" content="summary" />
+
+  {% if page.image %}
+    <meta name="twitter:card" content="summary_large_image" />
+  {% elsif  %}
+    <meta name="twitter:card" content="summary" />
+  {% endif %}
+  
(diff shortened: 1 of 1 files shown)


lib/template.html
```diff
@@ -124,7 +124,13 @@
 {% endif %}
 
 {% if site.twitter %}
-  <meta name="twitter:card" content="summary" />
+
+  {% if page.image %}
+    <meta name="twitter:card" content="summary_large_image" />
+  {% elsif  %}
```
benbalter: Believe this should be an `else`. Also, there's an extra trailing space.

DirtyF:
closing this in favor of #72

pathawks:
Sad day. I was hoping that, since this was merged via #72, the icon would turn purple instead of red. I guess because of 049cb11 GitHub thinks that it wasn't totally merged?

Anyway, the changes have made their way to **master**. Thank you for your efforts @DirtyF :beers:

DirtyF:
@pathawks BTW I really respect the work you're doin' here with @benbalter to help people contrib to Jekyll. I still fear I will not be able to help the projet, due to my poor development skills, and it's really great to see how you encourage people, beginners included. :tophat: (chapeau as we say in French).

PR #76, opened 2016-03-21
Add support for page images with absolute URLs

pathawks:


benbalter:
A feature that absorbs complexity for users, accounts for them not reading the manual, and does so without adding a setting?!?! 💖

lib/template.html
```diff
@@ -85,11 +85,19 @@
...
+  {% unless seo_site_logo contains "://" %}
+    {% assign seo_site_logo = seo_site_logo | prepend: seo_url %}
+  {% endunless %}
+  {% assign seo_site_logo = seo_site_logo | escape %}
 {% endif %}
 
 {% if page.image %}
-    {% assign seo_page_image = page.image | prepend: seo_url | escape %}
+  {% assign seo_page_image = page.image %}
+  {% unless seo_page_image contains "://" %}
```
DirtyF: contains "//" in case of prefix-free URLs like  `//jekyllrb.com/img/logo-2x.png` ?

pathawks: http://www.paulirish.com/2010/the-protocol-relative-url/

> Now that SSL is encouraged for everyone and doesn’t have performance concerns, *this technique is now an anti-pattern.* If the asset you need is available on SSL, then *always* use the `https://` asset.

pathawks:
I am sure this conflicts with #78.

benbalter:
@pathawks I just merged in https://github.com/jekyll/jekyll-seo-tag/pull/78 to get out of your way. Looking at the diff, I *believe* it should just be a one line conflict. But feel free to clobber my changes as you see fit, and merge when you're ready. Once you do, glad to cut a minor release.

pathawks:
Go team 🎉 

@jekyllbot: merge

pathawks:
@jekyllbot: merge

lib/template.html
```diff
@@ -85,11 +85,20 @@
...
+  {% unless seo_site_logo contains "://" %}
+    {% assign seo_site_logo = seo_site_logo | prepend: seo_url %}
+  {% endunless %}
+  {% assign seo_site_logo = seo_site_logo | escape %}
 {% endif %}
 
 {% if page.image %}
-    {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image | prepend: seo_url | escape %}
+  {% assign seo_page_image = page.image.path | default: page.image.facebook |
+default: page.image %}
```
benbalter: @pathawks is this line break intentional? Does it matter?

pathawks: In the process of changing text editors; didn't notice it was hard wrapping this :rage2:

pathawks: Fixed via #79

benbalter: > In the process of changing text editors

Hopefully to Atom! 😄

pathawks: > Hopefully to Atom

The other direction, actually 😞

Issue #77, opened 2016-04-19
Ability to support Facebook and Twitter specific images, set image height and width


benbalter:
Via @leereilly, Facebook and Twitter require different sized images, and for Facebook, if height and width is passed, it saves Facebook from making an async call, and thus the image loads faster. Isn't the internet a wonderful place?

I'm thinking that the `image` property should continue to work as it does today, if it's passed a string, e.g., `image: /foo.png`.

However, if the `image` property is an object (the "advanced usage"), we can look to and output the following (optional) properties:

* `path` - same as passing a string, but allows you to also specify height and width
* `twitter` - the path to the Twitter image
* `facebook` - the path to the facebook image
* `height` - the explicit image height 
* `width` - the explicit image width

PR #78, opened 2016-04-19
Ability to support Facebook and Twitter specific images, set image height and width

benbalter:
This PR should preserve the existing behavior, while allowing users to opt-in to more advanced control, should they need it.

Fixes https://github.com/jekyll/jekyll-seo-tag/issues/77. 

/cc @leereilly

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -89,7 +89,7 @@
 {% endif %}
 
 {% if page.image %}
-    {% assign seo_page_image = page.image | prepend: seo_url | escape %}
+    {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image | prepend: seo_url | escape %}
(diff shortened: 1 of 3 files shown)


lib/template.html
```diff
@@ -115,7 +115,17 @@
...
+<meta property="og:image" content="{{ seo_page_image }}" />
+  {% if page.image.height %}
+    <meta property="og:image:height" content="{{ page.image.height }}" />
+  {% endif %}
+  {% if page.image.width %}
+    <meta property="og:image:width" content="{{ page.image.width }}" />
+  {% endif %}
+{% endif %}
+
+{% if page.image.twitter %}
```
stereobooster: `{% if seo_page_image or page.image.twitter %}` bug here

PR #87, opened 2016-05-12
Different philosophy for rel prev/next links based on recommendations from Google

### LINKED ISSUE #80: Prev/next links are more appropriate for pagination?
Quote from [this google blogpost](https://webmasters.googleblog.com/2011/09/pagination-with-relnext-and-relprev.html):

> Throughout the web, a paginated series of content may take many shapes—it can be an article divided into several component pages, or a product category with items spread across several pages, or a forum thread divided into a sequence of URLs. 

I am not sure that it is good to use prev/next meta links on posts because content on each of them may be completely different and not related at all. However, I am sure that these links must be enabled on pages with pagination. What do you think?

tfe:
Details here: https://webmasters.googleblog.com/2011/09/pagination-with-relnext-and-relprev.html.

This replaces the per-page prev/next links with general ones that will be output whenever there is a paginator with a previous or next page to show. I like the idea of leaning on the paginator to know when there is a previous or next page to link to.

Also resolves #80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Jekyll SEO Tag adds the following meta tags to your site:
 * Pages title (with site title appended when available)
 * Page description
 * Canonical URL
-* Next and previous URLs for posts
+* Next and previous URLs on paginated pages
 * [JSON-LD Site and post metadata](https://developers.google.com/structured-data/) for richer indexing
 * [Open graph](http://ogp.me/) title, description, site title, and URL (for Facebook, LinkedIn, etc.)
 * [Twitter summary card](https://dev.twitter.com/cards/overview) metadata
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -30,6 +30,7 @@ module Jekyll
       {
         'page'    => context.registers[:page],
         'site'    => context.registers[:site].site_payload['site'],
+        'paginator' => context['paginator'],
         'seo_tag' => options
       }
     end
--- a/lib/template.html
+++ b/lib/template.html
@@ -139,14 +139,13 @@
 {% if page.date %}
   <meta property="og:type" content="article" />
   <meta property="article:published_time" content="{{ page.date | date_to_xmlschema }}" />
+{% endif %}
 
-  {% if page.next.url %}
-    <link rel="next" href="{{ page.next.url | prepend: seo_url | replace:'/index.html','/' }}" title="{{ page.next.title | escape }}" />
-  {% endif %}
-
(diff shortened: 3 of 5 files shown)


tfe:
Oops, this was intended for jekyll-seo-tag.

pathawks:
Looks like `next` and `prev` should only be used for pagination and not stringing every article into a chronological chain. Cool, thanks for catching this :+1:

Before this can be merged, we will need to add some tests. Are you able to add some specs to `spec/jekyll_seo_tag_spec.rb`? Since I think we can rely on Paginator to do the right thing, I think mocking up some data and checking for the expected output would be adaquate :+1:

pathawks:
Apparently we had no tests for the old way, since removing it didn't break anything 😶

tfe:
Yes, that's why I didn't write any for this change...

I'll see if I can find some time for it tomorrow.

pathawks:
I would love to merge this. @tfe will you have some time to write some tests?

tfe:
It took quite a bit of trial-and-error and poking around to figure out how to correctly stub the paginator for testing, but there it is. :)

tfe:
Oops, stand by.

tfe:
OK, green now. So I'm sure you thought of this, but this is a breaking change to the existing pagination behavior probably worthy of a version number bump.

benbalter:
> this is a breaking change to the existing pagination behavior probably worthy of a version number bump

@tfe could you elaborate a bit? Does it break configuration at all, or just changes the output?

pathawks:
@benbalter It only changes the output, not the interface.

@tfe Thinking a bit more about this, I think one could make a strong argument that this is merely a bugfix rather than a breaking change.

That said, I think this will be a big enough change that we need to very clearly communicate what is happening and why. I guess I could go either way as far as major/minor version bump.

benbalter:
I agree, we can call this a bug fix, but am fine making this a minor bump if we think it warrants it.

tfe:
Doesn't matter to me; I have no skin in the game. I just wanted to make sure you and the other maintainers thought about it before hitting merge.

benbalter:
LGTM. @pathawks do you think it needs a major bump, or is a minor bump sufficient? (I'd even be fine with calling it a bug fix, as long as we communicate the change in behavior). I can't see how this would break any site, even if it is a change in behavior.

pathawks:
Lets do a minor bump. This is bigger than just a patch release because it changes the behavior in a way we need to be sure to communicate, but no big API changes :+1:

benbalter:
Thanks @tfe! Nice work.

PR #90, opened 2016-05-20
Use SmartyPants

### LINKED ISSUE #17: Poor Man’s Smartypants

pathawks:
Fixes #17

Requires Jekyll 3.1

The `smartify` filter [includes Kramdown's `html_entity` parser](https://github.com/jekyll/jekyll/blob/v3.1.6/lib/jekyll/converters/smartypants.rb#L5) which will escape HTML for us. :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,3 +183,7 @@ image:
   height: 100
   width: 100
 ```
+
+### SmartyPants Titles
+
+Titles will be processed using [Jekyll's `smartify` filter](https://jekyllrb.com/docs/templates/). This will use SmartyPants to translate plain ASCII punctuation into "smart" typographic punctuation. This will not render or strip any Markdown you may be using in a page title.
--- a/lib/template.html
+++ b/lib/template.html
@@ -34,19 +34,19 @@
     {% assign seo_name = seo_site_title %}
 {% endif %}
 {% if seo_name %}
-  {% assign seo_name = seo_name | markdownify | strip_html | strip_newlines | escape_once %}
+  {% assign seo_name = seo_name | smartify | strip_html | strip_newlines | escape_once %}
 {% endif %}
 
 {% if seo_title %}
-  {% assign seo_title = seo_title | markdownify | strip_html | strip_newlines | escape_once %}
+  {% assign seo_title = seo_title | smartify | strip_html | strip_newlines | escape_once | escape_once %}
 {% endif %}
 
 {% if seo_site_title %}
-  {% assign seo_site_title = seo_site_title | markdownify | strip_html | strip_newlines | escape_once %}
+  {% assign seo_site_title = seo_site_title | smartify | strip_html | strip_newlines | escape_once %}
 {% endif %}
 
 {% if seo_page_title %}
-  {% assign seo_page_title = seo_page_title | markdownify | strip_html | strip_newlines | escape_once %}
+  {% assign seo_page_title = seo_page_title | smartify | strip_html | strip_newlines | escape_once %}
 {% endif %}
 
 {% assign seo_description = page.description | default: page.excerpt | default: site.description %}
--- a/jekyll-seo-tag.gemspec
+++ b/jekyll-seo-tag.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.add_dependency 'jekyll', '~> 3.0'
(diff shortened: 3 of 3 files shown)


lib/template.html
```diff
@@ -34,24 +34,27 @@
...
 {% if seo_site_title %}
-  {% assign seo_site_title = seo_site_title | markdownify | strip_html | strip_newlines | escape_once %}
+  {% assign seo_site_title = seo_site_title | smartify | strip_html |
+strip_newlines %}
 {% endif %}
 
 {% if seo_page_title %}
-  {% assign seo_page_title = seo_page_title | markdownify | strip_html | strip_newlines | escape_once %}
+  {% assign seo_page_title = seo_page_title | smartify | strip_html |
+strip_newlines %}
```
pathawks: STUPID VIM AND YOUR STUPID HARD BREAKS!

benbalter:
Right now, users can include markdown in e.g., the description (which would be rendered and stripped). Do we need to leave the `markdownify` filter in?

pathawks:
> Right now, users can include markdown in e.g., the description (which would be rendered and stripped).
> Do we need to leave the markdownify filter in?

We only need one or the other.

Is the description the only one where it makes sense to allow Markdown (which is stripped)? Can we assume that page title and site name should never be Markdown?

I know that there has been trouble in the past with titles rendered as Markdown that should not be. If my page title is `#hashtag`, the Markdown processor will convert it to `<h1>hashtag</h1>`, and when HTML is stripped it will leave `hashtag` which is less than ideal.

benbalter:
It'd be a change that we should document and be purposeful about. I guess I have two questions then:

1. Does `smartify` provide any benefit that `markdownify` does not?

2. Do we want to allow rendered, stripped markdown in the fields or leave it unrendered?

pathawks:
> Does `smartify` provide any benefit that `markdownify` does not?

The benefit of `smartify` is for text that was never intended to be processed via Markdown (ie: `#hashtag`).

Where might the description be used in a site template? I cannot think of a reason for the description to have any Markdown, especially if it is only used in `<meta>`

benbalter:
Makes sense. We'll want to make a note in the changelog, because users do the darnedest things, but otherwise 👍. Glad this finally made its way full circle.

pathawks:
Perhaps this should be a **2.0.0** feature, in case somebody relied on having markdown stripped from the title or something?

benbalter:
> Perhaps this should be a 2.0.0 feature, in case somebody relied on having markdown stripped from the title or something?

Makes sense. Perhaps this and https://github.com/jekyll/jekyll-seo-tag/pull/87 could go into a 2.0? Don't think there are any other breaking changes on the radar.

README.md
```diff
@@ -183,3 +183,7 @@ image:
   height: 100
   width: 100
 ```
+
+### SmartyPants Titles
+
+Titles will be processed using [Jekyll's `smartify` filter](https://jekyllrb.com/docs/templates/). This will use SmartyPants to translate plain ASCII punctuation into "smart" typographic punctuation. This will not render or strip any Markdown you may be using in a page title.
```
pathawks: Is this enough of an explanation?

pathawks:
I decided to only process titles, as it seemed a `description` might be more likely to contain Markdown?

benbalter:
>  as it seemed a description might be more likely to contain Markdown?

That makes sense, especially if the description is an auto-generated excerpt.

benbalter:
@pathawks this looks good to me. If you're happy with it and merge, glad to cut a 2.0 and get it up on Pages (we'd like to move the electron site over to [use Jekyll seo tag](https://github.com/electron/electron.atom.io/pull/248), and https://github.com/jekyll/jekyll-seo-tag/pull/87 is a blocker).

pathawks:
@jekyllbot: merge

PR #98, opened 2016-07-09
Use image for twitter unless twitter specific image provided

stereobooster:
If there is `seo_page_image` value, twitter card is `summary_large_image`, but the same time it is not providing tag for image

``` liquid
  {% if seo_page_image or page.image.twitter %}
    <meta name="twitter:card" content="summary_large_image" />
  {% else %}
    <meta name="twitter:card" content="summary" />
  {% endif %}
```


benbalter:
I believe this is the current behavior in `2.3.1`. Please comment if that is not the case and we can re-open/merge.

Issue #96, opened 2016-06-19
Add meta name="author"


renshuki:
It seems:
`<meta name="author" content="">`
is not added even if an author is specified in YAML front matter of a post/page/document.

Maybe useful to add?

pathawks:
Do you have any information on how this tag would be used (ie: by search engines)?

benbalter:
FWIW, there's also `rel=author` which Google uses for Google+ profile association.

benbalter:
> Is that still of any value?

Guess not. Good find.

renshuki:
Well, it's mostly used by Facebook when you share a blog article (By: author_name).  

See below:  
![image](https://cloud.githubusercontent.com/assets/7076736/17242355/ac229fc0-55b1-11e6-8ca3-7173fe650a52.png)

CGarces:
I have used https://developers.facebook.com/tools/debug/ to test the author tags.

My pages use a variation of the jekyll-seo-tag and contains: 

- Author and publisher at ld+json
- Meta property="article:publisher"
- Also one <a rel="author" >

Facebook debug tool not show the author in the preview.

After add `<meta name="author" content="">` the author is show in the preview..

benbalter:
Simple enough to add, if anyone is interested in submitting a pull request.

DirtyF:
@CGarces can you try with `<meta property="article:author" content="" />` and tell me if it works?  I noticed that SEO blogs use rather this meta.

CGarces:
Hi @DirtyF 
According to my results at https://developers.facebook.com/tools/debug/
`<meta property="article:author" content="" />` not is recognised by Facebook.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

jhabdas:
FWIW I just ran some tests using the Facebook debug tool and here were my results, confirmed by multiple runs with https://developers.facebook.com/tools/debug/:

**Test setup**
Ngrok-generated URLs submitted to debugger for the first time

[...]

PR #103, opened 2016-07-31
Add author meta

renshuki:
Hi,
Here is the fix for #96 
Thanks ;)

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -52,6 +52,15 @@
 
 {% assign seo_author = page.author | default: page.authors[0] | default: site.author %}
 {% if seo_author %}
+  {% if seo_author.name %}
+    {% assign seo_author_name = seo_author.name %}
+  {% else %}
+    {% if site.data.authors and site.data.authors[seo_author] %}
+      {% assign seo_author_name = site.data.authors[seo_author].name %}
+    {% else %}
+      {% assign seo_author_name = seo_author  %}
+    {% endif %}
+  {% endif %}
   {% if seo_author.twitter %}
     {% assign seo_author_twitter = seo_author.twitter %}
   {% else %}
@@ -104,8 +113,8 @@
   <meta property="og:title" content="{{ seo_page_title }}" />
 {% endif %}
 
-{% if seo_author %}
-  <meta name="author" content="{{ seo_author }}" />
+{% if seo_author_name %}
+  <meta name="author" content="{{ seo_author_name }}" />
 {% endif %}
 
 {% if seo_description %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -104,6 +104,10 @@
   <meta property="og:title" content="{{ seo_page_title }}" />
 {% endif %}
 
+{% if seo_author %}
+  <meta name="author" content="{{ seo_author }}" />
+{% endif %}
+
 {% if seo_description %}
   <meta name="description" content="{{ seo_description }}" />
   <meta property="og:description" content="{{ seo_description }}" />


benbalter:
LGTM. @pathawks?

pathawks:
Have we found any documentation regarding how this information is used?

renshuki:
Actually nothing in Facebook official documentation.  
- article:author   
- article:publisher  
must be an URL to a Facebook profile (or ID) to work properly.
  
In case those to tags are not defined, Facebook seems to use meta name="author".
So, even if you haven't a Facebook account you still can show up the author name of the article you share...

renshuki:
Little "UP" on this PR before closing. No needs?

pathawks:
I am :-1: without some documentation somewhere about how this works.

This is certainly a spa

renshuki:
Thanks for the update!
As I explained above nothing written explicitly in Facebook documentation.

> article:author  

only seems supported officially by Facebook but it requires a Facebook profile which is not suitable for everyone use case.

Some topics are dealing with the subject:

_Source: Stack Overflow_

If you want some official documentation proof, wait and see.

pathawks:
Fixes #96

jhabdas:
FYI on use of `meta name=author` in combination with `meta name=article:author` and how Facebook treats each of them differently: https://github.com/jekyll/jekyll-seo-tag/issues/96#issuecomment-275840637

jhabdas:
@renshuki I added info on how this works in my last comment. Think you could update the README and add some unit tests? If you need help just LMK.

lib/template.html
```diff
@@ -108,6 +108,10 @@
   <meta property="og:title" content="{{ seo_page_title }}" />
 {% endif %}
 
+{% if seo_author %}
+  <meta name="author" content="{{ seo_author }}" />
+{% endif %}
```
jhabdas: In the case `seo_author` is a key which maps to an item in `site.data.authors.yml` we should probably make the author name output using `site.data.authors[seo_author].name`.

renshuki: Great idea! I made the change in the last commit.

renshuki:
@jhabdas For unit tests, `site.data.authors` is actually in `context "twitter"` but the current feature is not related to Twitter so... Do you think a refactoring of the specs is preferable?

jhabdas:
@renshuki sorry for the delayed reply. A refactor makes to me if it helps untangle namespaces / contexts. May want to hear from a project maintainer first though so you don't lose any time. I think @pathawks may be looking for Facebook to document the feature (assuming existing documentation cannot be found).

renshuki:
@jhabdas Sure. I would greatly appreciate if Facebook could document this somewhere...

jhabdas:
@renshuki IMHO it doesn't matter if they document it. The fact it works at all (as we've shown) and adds standards-based metadata enriching the page should be justification enough for the addition.

pathawks:
> and adds standards-based metadata 

Where is that standard documented?

jhabdas:
http://devdocs.io/html/element/meta

jhabdas:
And here: https://www.w3.org/TR/html5/document-metadata.html#the-meta-element

pathawks:
You're right 👍🏼

benbalter:
Thanks @Thanks @renshuki @jhabdas! Merged via command line to confirm that rebasing allowed rubocop tests to pass.

jhabdas:
w00t

renshuki:
Thank you everyone 😄 
I will add some unit tests later for this if you want.

benbalter:
> I will add some unit tests later for this if you want.

A yes. Even if just to confirm the field is there/filled out to prevent a regression.

jhabdas:
@renahuki there lots of examples in the code you can learn from to get started. give it a shot! :)

PR #109, opened 2016-08-11
Inherit Jekyll's rubocop config for consistency

benbalter:
This PR is the result of adding `inherit_gem:\n jekyll: .rubocop.yml` to `.rubocop.yml` and running `rubocop -a` so that this project follows Jekyll core's coding styles for consistency like other core plugins. This way, as Jekyll's Ruby style changes, so too will this project's.

PR #119, opened 2016-09-08
Removes description from title in template

bnd5k:
This PR removes code that joined the site description to the title.

This issue is that search engines expect the title meta tag to be in the 50-60 character range  ([https://moz.com/learn/seo/title-tag](https://moz.com/learn/seo/title-tag)) and the description meta tag to be out 155 characters ([https://moz.com/learn/seo/meta-description](https://moz.com/learn/seo/meta-description)).   

If you join them, you end up with a title that's in the 205-215 characters range.

I think the default behavior should be to  keep the description and title separate.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -21,9 +21,6 @@
   {% assign seo_title = seo_site_title %}
   {% assign seo_page_title = seo_site_title %}
 
-  {% if site.description %}
-    {% assign seo_title = seo_title | append:" - " | append: site.description %}
-  {% endif %}
 {% endif %}
 
 {% if page.seo and page.seo.name %}
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -26,7 +26,8 @@ def make_page(options = {})
 end
 
 def make_post(options = {})
-  filename = File.expand_path('2015-01-01-post.md', CONFIG_DEFAULTS['source'])
+  #spec/fixtures/2015-01-01-post.md
+  filename = File.expand_path('2015-01-01-post.md', "#{CONFIG_DEFAULTS['source']}/_posts")
   config = { :site => site, :collection => site.collections['posts'] }
   page = Jekyll::Document.new filename, config
   page.merge_data!(options)


bnd5k:
This build is failing due to preexisting rubocop violations.  There are 273 violations that are not related to the code I removed.  

Can we consider this a passing build and then have a separate discussion about what to do with the rubocop violations?  (i.e. configure rubocop to only evaluate the changed code in a PR or dig in and fix all 273 rubocop violations)

pathawks:
I don't understand why this is failing but #109 didn't. @benbalter?

pathawks:
I don't have much of an opinion either way on this.

benbalter:
> If you join them, you end up with a title that's in the 205-215 characters range.

This PR assumes that users will:

1.  Have a title + description > 60 characters, and 
2. That exceeding that number would result in an SEO hit

For the first, [on my own site](http://ben.balter.com) as an example, my title + description is 61 characters, and it displays fine on Google's search results.

For the second, it sounds like those numbers are the optimal, based on what's commonly displayed in search engines, but there's no downside in terms of machine, SEO, other than readability. That sounds to me like an issue of optimizing the description, no?

In terms of scope, IIRC, this should only affect pages without a description, correct?

benbalter:
> This build is failing due to preexisting rubocop violations. https://github.com/jekyll/jekyll-seo-tag/commit/2631e3ea77c5863fbc81f55222fdf3c06eda0f88

bnd5k:
In terms of scope, the current functionality of combining the title and description only affects pages **with** descriptions.  So if you have a description, it's automatically duplicated inside your title. 

I don't see  the value of duplicating the description inside the title tag.  Maybe I'm missing something?

More importantly, a user should be able to set a title and a description, without having to worry that the gem is going to mutate  those values.

bnd5k:
This gem is awesome, BTW.  Thanks for maintaining it.  I'm sure it's not always fun to have random PR's throws your way without warning.

benbalter:
@bnd5k The intended functionality is:

* If the page has a no title: `SITE TITLE - SITE DESCRIPTION` (e.g., the home page)
* If the page has a title `PAGE TITLE - SITE TITLE` (e.g., any other page)

Is that not the case? If so, that's a bug from the original intention.

Looking at my own site, which uses the plugin:

* http://ben.balter.com/ - "Ben Balter - Attorney, open source developer, product manager"
* http://ben.balter.com/about/ - "About - Ben Balter"

bnd5k:
Thanks for your comments. I was confused about the interaction between the site title and page title.  

Everything here is working as expected.  Totally my fault for misunderstanding the source code.

Closing this PR.  Thanks for your help on this.

benbalter:
No worries. Thanks for being critical.  :smile:

PR #124, opened 2016-10-12
Add seo_description to twitter card.

enzanki-ars:
See https://dev.twitter.com/cards/types/summary for reason for change.  `twitter:description -> required=true`


enzanki-ars:
I'm currently learning about how the test code works, but until then, no testing code has been added.

pathawks:
Twitter will [use `og:description` if present](https://dev.twitter.com/cards/markup), so this is unnecessary.

Thank you for the PR though :+1:

PR #125, opened 2016-10-15
Simplify minify regex

pathawks:
This simplifies the minification regular expression.

I _did_ change one test, as the JSON-LD lines are no longer indented, although each attribute is still on its own line. Of course, if we would like to further minify things we can strip out _all_ the newlines, but that might negatively impact readability of the source and would not really gain anything (especially if we assume pages will be served GZIPed).


pathawks:
@jekyllbot: merge

_Pretty Please?_

PR #131, opened 2016-10-23
Allow to specify a default image in config file

### LINKED ISSUE #130: Can image information _config.yml be utilized if not present in front matter?
>The SEO tag will respect any of the following if included in your site's `_config.yml` (and simply not include them if they're not defined):

>* `title` - Your site's title (e.g., Ben's awesome site, The GitHub Blog, etc.)
>* `description` - A short description (e.g., A blog dedicated to reviewing cat gifs)
>* `url` - The full URL to your site. Note: `site.github.url` will be used by default.
>* `author` - global author information (see below)
>* `twitter:username` - The site's Twitter handle. You'll want to describe it like so:

>The SEO tag will respect the following YAML front matter if included in a post, page, or document:

>* `title` - The title of the post, page, or document
>* `description` - A short description of the page's content
>* `image` - URL to an image associated with the post, page, or document (e.g., `/assets/page-pic.jpg`)
>* `author` - Page-, post-, or document-specific author information (see below)

---

`title`, `description`, `author` if not present in YAML front matter, fallbacks to information available in `_config.yml`. 

Wondering, if `image`  can/should be treated in the same way?

DirtyF:
Fix #130 

/cc @pathawks @benbalter

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -88,8 +88,8 @@
   {% assign seo_site_logo = seo_site_logo | escape %}
 {% endif %}
 
-{% if page.image %}
-  {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image %}
+{% if page.image or site.image %}
+  {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image | default: site.image %}
   {% unless seo_page_image contains "://" %}
     {% assign seo_page_image = seo_page_image | absolute_url %}
   {% endunless %}


pathawks:
I would much rather use front matter defaults for this purpose, instead of reinventing the wheel.

DirtyF:
@pathawks What a fool am I, you're totally right, we just need to add an example in the documentation.

PR #132, opened 2016-10-23
Add front matter default example for image

### LINKED ISSUE #97: Document how to use front matter defaults
is it possible to set a default image for the `image:` frontmatter? would be great to be able to set a default (or you might call it fallback) setting for images. It just does not feel right to add the same image to all posts/articles

using

```
image: <the image value>
```

does not change any page at the moment

DirtyF:
Fix #97 

Would this be enough to start with?

/cc @benbalter @pathawks

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,20 @@ image:
   width: 100
 ```
 
+### Setting a default image
+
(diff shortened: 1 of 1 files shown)


benbalter:
@jekyllbot: merge +docs

benbalter:
Thanks @DirtyF!

PR #148, opened 2017-01-12
Add support for Bing and Yandex webmaster tools. Closes #147

jhabdas:
Adds value. Ideally would only appear on the front page as suggested by @pathawks in #45.

PR #149, opened 2017-01-13
Rubocop shouldn't worry if tests are too long

pathawks:
Jekyll [does the same thing](https://github.com/jekyll/jekyll/blob/v3.3.1/.rubocop.yml#L21-L24)

/cc #148


pathawks:
@jekyllbot: merge +dev

PR #151, opened 2017-01-16
Add tags to JSON metadata

aav7fl:
- [x]  Add Feature
- [X]  Add Docs
- [X]  Add Tests

This is just to add in the Author field (which is helpful for blog posts/articles when searching for articles + rich text formatting from Google). 

This also allows the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool) to pass without any errors for the type **BlogPosting** (now there are only warnings). 

BlogPosting was chosen instead of WebSite because of Google will treat this content type differently when it displays [Rich Cards](https://developers.google.com/search/docs/guides/search-gallery).

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,10 +215,9 @@ Or if you have no site specific images simply:
 ```yml
 image:
   path: /img/banner.png
-  height: 100
-  width: 100
 ```
 
+> Using the image dimensions are optional, but the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) will consider the JSON-LD to have errors when an image does not have specified dimensions.
 
 ### Setting a default image
 
--- a/README.md
+++ b/README.md
@@ -185,27 +185,25 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
-  * `dateModified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified. 
-  * `author` - The `author` field will match the JSON-LD author output where applicable.
+  * `date_modified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified.
 
 ### Customizing image output
 
 For most users, setting `image: [path-to-image]` on a per-page basis should be enough. If you need more control over how images are represented, the `image` property can also be an object, with the following options:
 
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
-* `url` - The path to the image to be used in the JSON-LD image object `url`.
 * `twitter` - The relative path to a Twitter-specific image.
 * `facebook` - The relative path to a Facebook-specific image.
 * `height` - The height of the Facebook (`og:image`) image and JSON-LD image object.
 * `width` - The width of the Facebook (`og:image`) image and JSON-LD image object.
 
-The JSON-LD will default to the `image: ` tag unless `url: ` `height: ` and `width: `.
+The JSON-LD will default to the `image: ` tag unless `path: ` `height: ` and `width: `.
 
 You can use any of the above, optional properties, like so:
 
 ```yml
 image:
-  url: /img/banner.png
+  path: /img/banner.png
   twitter: /img/twitter.png
   facebook: /img/facebook.png
   height: 100
@@ -216,7 +214,7 @@ Or if you have no site specific images simply:
 
 ```yml
 image:
-  url: /img/banner.png
+  path: /img/banner.png
   height: 100
   width: 100
 ```
--- a/README.md
+++ b/README.md
@@ -221,6 +221,17 @@ image:
 
 Using the image dimensions are optional, but the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) will consider the JSON-LD to have errors with certain types (i.e. BlogPosting) when an image does not have specified dimensions.
 
+#### Deprecated image output
+
+The older method for setting facebook image dimensions (shown below) is **no longer recommended**. It will continue work as intended, but is no longer the suggested procedure due to the lack of clarity in where these dimensions are assigned. 
+
+```yml
+image:
+  facebook: /img/bad_facebook_example.png		
+  height: 100		
+  width: 100		
+```
+
 ### Setting a default image
 
 You can define a default image using [Front Matter default](https://jekyllrb.com/docs/configuration/#front-matter-defaults), to provide a default Twitter Card or OGP image to all of your posts and pages.
--- a/README.md
+++ b/README.md
@@ -198,8 +198,6 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
 * `height` - The height of the image in pixels.
 * `width` - The width of image in pixels.
 
-The JSON-LD will default to the `image: ` tag unless `path: ` `height: ` and `width: `.
-
 You can use any of the above, optional properties, like so:
 
 ```yml
--- a/README.md
+++ b/README.md
@@ -78,8 +78,7 @@ The SEO tag will respect any of the following if included in your site's `_confi
 * `social` - For [specifying social profiles](https://developers.google.com/structured-data/customize/social-profiles). The following properties are available:
   * `name` - If the user or organization name differs from the site's name
   * `links` - An array of links to social media profiles.
-  * `date_modified` - (YYYY-MM-DD hh:mm) A manual override for the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` output. This is useful when the file timestamp does not match the true time that the content was modified.
-  * A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
+  * `date_modified` - Manually specify the `dateModified` field in the JSON-LD output to override Jekyll's own `dateModified`. This field will take **first priority** for the `dateModified` JSON-LD output. This is useful when the file timestamp does not match the true time that the content was modified. A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
 
   ```yml
   social:
--- a/README.md
+++ b/README.md
@@ -185,26 +185,42 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
+  * `dateModified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified. 
+  * `author` - The `author` field will match the JSON-LD author output where applicable.
 
 ### Customizing image output
 
 For most users, setting `image: [path-to-image]` on a per-page basis should be enough. If you need more control over how images are represented, the `image` property can also be an object, with the following options:
 
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
+* `url` - The path to the image to be used in the JSON-LD image object `url`.
 * `twitter` - The relative path to a Twitter-specific image.
 * `facebook` - The relative path to a Facebook-specific image.
-* `height` - The height of the Facebook (`og:image`) image
-* `width` - The width of the Facebook (`og:image`) image
+* `height` - The height of the Facebook (`og:image`) image and JSON-LD image object.
+* `width` - The width of the Facebook (`og:image`) image and JSON-LD image object.
+
+The JSON-LD will default to the `image: ` tag unless `url: ` `height: ` and `width: `.
 
 You can use any of the above, optional properties, like so:
 
 ```yml
 image:
+  url: /img/banner.png
   twitter: /img/twitter.png
   facebook: /img/facebook.png
   height: 100
   width: 100
 ```
+
+Or if you have no site specific images simply:
+
+```yml
+image:
+  url: /img/banner.png
+  height: 100
+  width: 100
+```
+
 
 ### Setting a default image
 
--- a/README.md
+++ b/README.md
@@ -185,7 +185,8 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
-  * `date_modified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified.
+  * `date_modified` - (YYYY-MM-DD hh:mm) A manual override for the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` output. This is useful when the file timestamp does not match the true time that the content was modified.
+    * A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
 
 ### Customizing image output
 
@@ -194,30 +195,33 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
 * `twitter` - The relative path to a Twitter-specific image.
 * `facebook` - The relative path to a Facebook-specific image.
-* `height` - The height of the Facebook (`og:image`) image and JSON-LD image object.
-* `width` - The width of the Facebook (`og:image`) image and JSON-LD image object.
+* `height` - The height of the image in pixels.
+* `width` - The width of image in pixels.
 
 The JSON-LD will default to the `image: ` tag unless `path: ` `height: ` and `width: `.
 
 You can use any of the above, optional properties, like so:
 
 ```yml
-image:
-  path: /img/banner.png
-  twitter: /img/twitter.png
-  facebook: /img/facebook.png
-  height: 100
-  width: 100
+image: /img/banner.png
 ```
 
-Or if you have no site specific images simply:
+Or if you have Facebook or Twitter specific images simply use:
 
 ```yml
 image:
-  path: /img/banner.png
+  default:
+    path: /img/banner.png
+    height: 100
+    width: 100
+  facebook:
+    path: /img/facebook.png
+    height: 90
+    width: 90
+  twitter: /img/twitter.png
 ```
 
-> Using the image dimensions are optional, but the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) will consider the JSON-LD to have errors when an image does not have specified dimensions.
+Using the image dimensions are optional, but the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) will consider the JSON-LD to have errors with certain types (i.e. BlogPosting) when an image does not have specified dimensions.
 
 ### Setting a default image
 
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ The SEO tag will respect any of the following if included in your site's `_confi
 * `social` - For [specifying social profiles](https://developers.google.com/structured-data/customize/social-profiles). The following properties are available:
   * `name` - If the user or organization name differs from the site's name
   * `links` - An array of links to social media profiles.
+  * `date_modified` - (YYYY-MM-DD hh:mm) A manual override for the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` output. This is useful when the file timestamp does not match the true time that the content was modified.
+  * A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
 
   ```yml
   social:
@@ -185,8 +187,6 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
-  * `date_modified` - (YYYY-MM-DD hh:mm) A manual override for the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` output. This is useful when the file timestamp does not match the true time that the content was modified.
-    * A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
 
 ### Customizing image output
 
@@ -195,41 +195,17 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
 * `twitter` - The relative path to a Twitter-specific image.
 * `facebook` - The relative path to a Facebook-specific image.
-* `height` - The height of the image in pixels.
-* `width` - The width of image in pixels.
+* `height` - The height of the Facebook (`og:image`) image
+* `width` - The width of the Facebook (`og:image`) image
 
 You can use any of the above, optional properties, like so:
 
 ```yml
-image: /img/banner.png
-```
-
-Or if you have Facebook or Twitter specific images simply use:
-
-```yml
 image:
-  default:
-    path: /img/banner.png
-    height: 100
-    width: 100
-  facebook:
-    path: /img/facebook.png
-    height: 90
-    width: 90
   twitter: /img/twitter.png
-```
-
-Using the image dimensions are optional, but the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) will consider the JSON-LD to have errors with certain types (i.e. BlogPosting) when an image does not have specified dimensions.
-
-#### Deprecated image output
-
-The older method for setting facebook image dimensions (shown below) is **no longer recommended**. It will continue work as intended, but is no longer the suggested procedure due to the lack of clarity in where these dimensions are assigned. 
-
-```yml
-image:
-  facebook: /img/bad_facebook_example.png		
-  height: 100		
-  width: 100		
+  facebook: /img/facebook.png
+  height: 100
+  width: 100
 ```
 
 ### Setting a default image
--- a/lib/template.html
+++ b/lib/template.html
@@ -115,8 +115,8 @@
     {% assign seo_page_image_facebook = seo_page_image_facebook | absolute_url %}
   {% endunless %}
   {% assign seo_page_image_facebook = seo_page_image_facebook | escape %}
-  {% assign seo_page_image_facebook_width = page.image.facebook.width %}
-  {% assign seo_page_image_facebook_height = page.image.facebook.height %}
+  {% assign seo_page_image_facebook_width = page.image.facebook.width | default: page.image.width %}
+  {% assign seo_page_image_facebook_height = page.image.facebook.height | default: page.image.height %}
 
   {% assign seo_page_image_twitter = page.image.twitter.path | default: page.image.twitter %}
   {% unless seo_page_image_twitter contains "://" %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -104,10 +104,6 @@
   <meta property="og:title" content="{{ seo_page_title }}" />
 {% endif %}
 
-{% if seo_author %}
-  <meta name="author" content="{{ seo_author }}" />
-{% endif %}
-
 {% if seo_description %}
   <meta name="description" content="{{ seo_description }}" />
   <meta property="og:description" content="{{ seo_description }}" />
--- a/lib/template.html
+++ b/lib/template.html
@@ -62,8 +62,8 @@
     {% endif %}
   {% endif %}
   {% assign seo_author_twitter = seo_author_twitter | replace:"@","" %}
-  {% if seo_author.name %}
-    {% assign seo_author_name = seo_author.name %}
+  {% if seo_author %}
+    {% assign seo_author_name = seo_author %}
   {% endif %}
 {% endif %}
 
--- a/lib/template.html
+++ b/lib/template.html
@@ -63,7 +63,7 @@
   {% endif %}
   {% assign seo_author_twitter = seo_author_twitter | replace:"@","" %}
   {% if seo_author %}
-    {% assign seo_author_name = seo_author %}
+    {% assign seo_author_name = seo_author.name | default: seo_author %}
   {% endif %}
 {% endif %}
 
@@ -210,7 +210,10 @@
 {% endif %}
 
 {% if seo_author_name %}
-    "author": {{ seo_author_name | jsonify }},
+    "author": {
+      "@type": "Person",
+      "name": {{ seo_author_name | jsonify }}
+    },
 {% endif %}
 
 {% if seo_page_image %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -62,9 +62,6 @@
     {% endif %}
   {% endif %}
   {% assign seo_author_twitter = seo_author_twitter | replace:"@","" %}
-  {% if seo_author %}
-    {% assign seo_author_name = seo_author.name | default: seo_author %}
-  {% endif %}
 {% endif %}
 
 {% if page.seo and page.seo.type %}
@@ -105,6 +102,10 @@
 
 {% if seo_page_title %}
   <meta property="og:title" content="{{ seo_page_title }}" />
+{% endif %}
+
+{% if seo_author %}
+  <meta name="author" content="{{ seo_author }}" />
 {% endif %}
 
 {% if seo_description %}
@@ -209,10 +210,10 @@
     "headline": {{ seo_page_title | jsonify }},
 {% endif %}
 
-{% if seo_author_name %}
+{% if seo_author %}
     "author": {
       "@type": "Person",
-      "name": {{ seo_author_name | jsonify }}
+      "name": {{ seo_author | jsonify }}
     },
 {% endif %}
 
@@ -246,8 +247,8 @@
 {% if seo_site_logo %}
     "publisher": {
         "@type": "Organization",
-        {% if seo_author_name %}
-            "name": {{ seo_author_name | jsonify }},
+        {% if seo_author %}
+            "name": {{ seo_author | jsonify }},
         {% endif %}
         "logo": {
             "@type": "ImageObject",
--- a/lib/template.html
+++ b/lib/template.html
@@ -92,7 +92,7 @@
 {% endif %}
 
 {% if page.image %}
-  {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image %}
+  {% assign seo_page_image = page.image.path | default: page.image.url | default: page.image.facebook | default: page.image %}
   {% unless seo_page_image contains "://" %}
     {% assign seo_page_image = seo_page_image | absolute_url %}
   {% endunless %}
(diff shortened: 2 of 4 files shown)


lib/template.html
```diff
@@ -221,6 +228,9 @@
 {% if seo_site_logo %}
     "publisher": {
         "@type": "Organization",
+        {% if seo_author_name %}
+            "Name": {{ seo_author_name | jsonify }},
```
pathawks: Shouldn't this be [`name`](http://schema.org/name)?

pathawks:
Does this fix #96?

/cc: @renshuki

lib/template.html
aav7fl: Yes. It should. I'll change that shortly.

aav7fl:
It does not fix #96. I can get around to that soon. It's something I would like to add as well. This only adds the proper JSON tags to pass the errors (but not the warnings).

pathawks:
> I can get around to that soon. It's something I would like to add as well.

I was just curious. We've got a PR for that in #103 :+1:

lib/template.html
```diff
@@ -89,7 +92,7 @@
 {% endif %}
 
 {% if page.image %}
-  {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image %}
+  {% assign seo_page_image = page.image.path | default: page.image.url | default: page.image.facebook | default: page.image %}
```
aav7fl: Allow for page.image.url here, else there will be a mess of if interpreting the yaml object when it encounters this unknown yaml field.

lib/template.html
```diff
@@ -62,6 +62,9 @@
     {% endif %}
   {% endif %}
   {% assign seo_author_twitter = seo_author_twitter | replace:"@","" %}
+  {% if seo_author.name %}
+    {% assign seo_author_name = seo_author.name %}
+  {% endif %}
```
aav7fl: Add Author Name

lib/template.html
```diff
@@ -206,28 +209,57 @@
     "headline": {{ seo_page_title | jsonify }},
 {% endif %}
 
+{% if seo_author_name %}
+    "author": {{ seo_author_name | jsonify }},
+{% endif %}
+
```
aav7fl: Add Author Name

lib/template.html
```diff
@@ -206,28 +209,57 @@
...
+    {% if page.image.height && page.image.width %}
+      "image": {
+          "@type": "ImageObject",
+          "url": {{ seo_page_image | jsonify }},
+          "height": {{ page.image.height | jsonify }},
+          "width": {{ page.image.width | jsonify }}
+      },
+    {% else %}
+        "image": {{ seo_page_image | jsonify }},
+    {% endif %}
```
aav7fl: Create image object (as it needs). Height and width are probably required now for some future thinking. Think AMP where they [require this information](https://www.ampproject.org/docs/reference/components/amp-img) in the markup. 

Will default back if either no width/height are present, or if no image url is present.

lib/template.html
```diff
@@ -206,28 +209,57 @@
...
 {% if page.date %}
     "datePublished": {{ page.date | date_to_xmlschema | jsonify }},
 {% endif %}
 
+{% if page.dateModified %}
+    "dateModified": {{ page.dateModified | date_to_xmlschema | jsonify }},
+{% elseif page.date %}
+    "dateModified": {{ page.date | date_to_xmlschema | jsonify }},
+{% endif %}
+
```
aav7fl: Add a date modified field. Useful when wanting to control this manually. Like when I make a change to the markup file, but I want to override the timestamp given.

Will default to page.date if never "modified". Will not exist if neither field present.

lib/template.html
```diff
@@ -206,28 +209,57 @@
...
 {% if seo_description %}
     "description": {{ seo_description | jsonify }},
 {% endif %}
 
 {% if seo_site_logo %}
     "publisher": {
         "@type": "Organization",
+        {% if seo_author_name %}
+            "name": {{ seo_author_name | jsonify }},
+        {% endif %}
```
aav7fl: Add Author Name.

lib/template.html
```diff
@@ -206,28 +209,57 @@
...
     },
 {% endif %}
 
+{% if seo_type == "BlogPosting" %}
+    "mainEntityOfPage": {
+         "@type": "WebPage",
+         "@id": {{ page.url | replace:'/index.html','/' | absolute_url | jsonify }}
+    },
+{% endif %}
+
```
aav7fl: Add mainEntityOfPage when BlogPosting type is chosen. 
Better explained in this [StackOverflow post](http://stackoverflow.com/a/34467088/1813592).

aav7fl:
Is see that it's failing some builds. I'll look into this right now.

aav7fl:
Added updated documentation for changes.

aav7fl:
- [X]  Add Feature
- [X]  Add Tests
- [X]  Add Docs

Travis didn't like my date/dateModified tests because of time-zone issues. I had to pull those specific ones. Fixed up the rubocop styling. Looks like it's good for an inspection/questioning.

This makes changes to the JSON-LD information. Fixes warning/errors reported by [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool?url).

lib/template.html
```diff
@@ -104,6 +104,10 @@
   <meta property="og:title" content="{{ seo_page_title }}" />
 {% endif %}
 
+{% if seo_author %}
+  <meta name="author" content="{{ seo_author }}" />
+{% endif %}
+
```
aav7fl: Added in content of PR #103 along with tests (because it made sense to add the author meta field when I was already doing it in the JSON).

pathawks: I am still :-1: on this with out a clear use case

DirtyF: meta author is not recognized by Google (https://support.google.com/webmasters/answer/79812?hl=en) et was deprecated in favor of Rich Snippets. What is the intent in adding if it is ignored by search engines?

aav7fl:
Ok. Yea. I guess I'm stuck in the past. I can't find any good documentation that clearly states that it is used anymore. Even for markup. Most places with use JSON-LD or OG tags. Removing author meta tag & test.

lib/template.html
renshuki: Well it's ignored by Google but not by Facebook. The purpose of this change is to show up the author name of an article when you share it on Facebook. If the `meta name="author"` is not set the author name will not be displayed when you share a jekyll blog article on Facebook (Issue #96). If the author of the article has a Facebook account you can use the `meta property="article:author"` which his ID / profile URL instead but it's not always the case.  

If someone have any other suggestion to resolve this problem it would be greatly appreciated :smiley:

README.md
```diff
@@ -185,27 +185,43 @@ The following options can be set for any particular page. While the default opti
...
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
+  * `dateModified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified. 
+  * `author` - The `author` field will match the JSON-LD author output where applicable.
 
 ### Customizing image output
 
 For most users, setting `image: [path-to-image]` on a per-page basis should be enough. If you need more control over how images are represented, the `image` property can also be an object, with the following options:
 
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
+* `url` - The path to the image to be used in the JSON-LD image object `url`.
```
jhabdas: IMHO we should not have both a `path` and a `url`. It's downright confusing.

jhabdas: It would be nice to have a human-readable description or link to additional information on why one might want to set the `JSON-LD` separately from `path`. Ideally the user would specify `image` or `image:path` and it everything would just work across the board.

README.md
```diff
@@ -185,27 +185,43 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
+  * `dateModified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified.
```
benbalter: This should be `snake_case` not titleCase to match Ruby conventions. See, e.g., `webmaster_verifications` below.

README.md
```diff
@@ -185,27 +185,43 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
+  * `dateModified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified. 
+  * `author` - The `author` field will match the JSON-LD author output where applicable.
```
benbalter: How does this differ from the author logic above?

README.md
benbalter: Agree with @jhabdas that we should be using the existing field here.

lib/template.html
```diff
@@ -206,28 +206,60 @@
...
 
 {% if page.date %}
     "datePublished": {{ page.date | date_to_xmlschema | jsonify }},
 {% endif %}
 
+{% if page.dateModified %}
+    "dateModified": {{ page.dateModified | date_to_xmlschema | jsonify }},
+{% elsif page.date %}
+    "dateModified": {{ page.date | date_to_xmlschema | jsonify }},
+{% endif %}
```
benbalter: Can you link to some documentation explaining how this information is used? Is there value in providing the modified date when it doesn't differ from the publishing date?

aav7fl: Sure. [Google's Docs](https://developers.google.com/search/docs/data-types/articles) mention that the `dateModified` field is recommended. I can't give an exact explanation for its use other than it receives Google's recommendation. 

Under this documentation, is listed as recommended for AMP pages, and ignored on non-AMP. But then again, it also says that `datePublished` is ignored on non-AMP pages. What are your thoughts?

README.md
aav7fl: It doesn't differ. It was documentation letting the user know that the author field was now being used in the JSON-LD. I can see how its placement would imply that it is an additional entry. I will remove this because the author field is already mentioned as a global property earlier in the documentation.

README.md
```diff
@@ -185,6 +185,7 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
+  * `date_modified` - An override for the `dateModified` field in the JSON-LD output. Useful when the file timestamp does not match the true time that the content was modified.
```
aav7fl: (Comment from an earlier commit)
Google's Docs mention that the dateModified field is recommended. I can't give an exact explanation for its use other than it receives Google's recommendation.

Under this documentation, is listed as recommended for AMP pages, and ignored on non-AMP. But then again, it also says that datePublished is ignored on non-AMP pages. What are your thoughts?

aav7fl:
Going to run some tests on this merge.

aav7fl:
Alright. Done checking on my personal build. What are the thoughts? It's been a while since there has been any activity on this PR from reviewers. I'd like to know if I can answer any questions or change things to match the project styling/goal.

README.md
```diff
@@ -193,19 +194,32 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
 * `twitter` - The relative path to a Twitter-specific image.
 * `facebook` - The relative path to a Facebook-specific image.
-* `height` - The height of the Facebook (`og:image`) image
-* `width` - The width of the Facebook (`og:image`) image
+* `height` - The height of the Facebook (`og:image`) image and JSON-LD image object.
+* `width` - The width of the Facebook (`og:image`) image and JSON-LD image object.
+
```
jhabdas: From a user standpoint this feels like TMI. As a user of the Gem, I just want to know it supports JSON-LD and not have to think about how it works.

README.md
```diff
@@ -193,19 +194,32 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
...
+The JSON-LD will default to the `image: ` tag unless `path: ` `height: ` and `width: `.
 
 You can use any of the above, optional properties, like so:
 
 ```yml
 image:
+  path: /img/banner.png
```
jhabdas: Is path allowed with these other values?

README.md
```diff
@@ -193,19 +194,32 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
...
 ```
 
+Or if you have no site specific images simply:
+
+```yml
+image:
+  path: /img/banner.png
+  height: 100
+  width: 100
+```
```
jhabdas: This is only if one wants to control size, correct? Otherwise they can just specify image as a path if I'm not mistaken. If that's true this is probably not needed as it may make the doc less approachable.

lib/template.html
```diff
@@ -219,28 +219,60 @@
...
 {% if seo_description %}
     "description": {{ seo_description | jsonify }},
 {% endif %}
 
 {% if seo_site_logo %}
     "publisher": {
         "@type": "Organization",
+        {% if seo_author %}
+            "name": {{ seo_author | jsonify }},
+        {% endif %}
```
jhabdas: I understand Google's Structured Data Testing Tool requires publisher to be an Organization while Schema.org spec allows Person. So I have to ask, should publisher as an Organization be a person's name, or the site name?

lib/template.html
jhabdas: I tried read that post 6 six times, studied Schema.org and still couldn't understand this until I saw your example. 😂 Given Schema.org says all webpages are implicitly of Type WebPage (if not defined) this makes sense. Given this tag also applies to "CreativeWork", it would be useful to add that here while we're making this change as it will help with Structured Data validation for that common type.

spec/jekyll_seo_tag_spec.rb
```diff
@@ -196,6 +196,60 @@
       end
     end
 
+    context "with image.path, image.height, and image.width" do
```
jhabdas: Seems this should be combined with existing test to help keep the logic DRY

README.md
aav7fl: I put path in here because it wasn't clear if I used "facebook" as a url if it would count as the `url` for the image in JSON.

README.md
aav7fl: I've found that if I don't include this, the Google Schema checker will throw an error (not a warning) to the JSON markup. It can't be generated _"correctly"_ because it can't be generated.

README.md
aav7fl: That is correct, but then the JSON-LD won't pass the Google Schema checker without the image dimensions.

lib/template.html
aav7fl: I suppose it depends how your website is created. A valid point.

spec/jekyll_seo_tag_spec.rb
aav7fl: Where might you suggest this goes? It didn't feel right placing it with the other "with image.path only" (Or around that) test because it looks like the tests are broken into `<meta>` and `json-ld` sections.

lib/template.html
aav7fl: Thanks! I've added this to my local branch, along with a test.

README.md
aav7fl: I will update the documentation to mention this, alter this example to remove the dimensions, and bring up the Google Structured Data Testing Tool.

README.md
jhabdas: Ick, really? I'd expect that to a nice to have for any kind of Structured Data. With microformats dimensions weren't required, just an image.

spec/jekyll_seo_tag_spec.rb
jhabdas: BDD FTW IMHO

jhabdas: In other words, feels like the smell is that schema goods are being duplicated by implementation details and not behavior. Seems like a good candidate for a letter refactor.

jhabdas:
Pat or Ben, a second pair of eyes? If this adds value let's ship it

spec/jekyll_seo_tag_spec.rb
aav7fl: Alright. Should we save the testing refactor for a separate PR? It seems like a lot of the tests need to be refactored. It appears fall outside of the scope that I was hoping to accomplish with this PR. I'm not against it, but it's going to get harder to refactor with other PRs getting accepted before this one with all of the changes being made.

jhabdas: If you're up for it, I feel like the template could see a 20% drop in LOC with refactoring. Then a separate refactoring of the tests.

jhabdas: (In a separate PR, totally)

README.md
benbalter: What format should the date be?

README.md
```diff
@@ -193,19 +194,31 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
 * `path` - The relative path to the image. Same as `image: [path-to-image]`
 * `twitter` - The relative path to a Twitter-specific image.
 * `facebook` - The relative path to a Facebook-specific image.
-* `height` - The height of the Facebook (`og:image`) image
-* `width` - The width of the Facebook (`og:image`) image
+* `height` - The height of the Facebook (`og:image`) image and JSON-LD image object.
+* `width` - The width of the Facebook (`og:image`) image and JSON-LD image object.
+
+The JSON-LD will default to the `image: ` tag unless `path: ` `height: ` and `width: `.
```
benbalter: What are you trying to say here?

README.md
```diff
@@ -193,19 +194,31 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
...
 ```
 
+Or if you have no site specific images simply:
+
+```yml
+image:
+  path: /img/banner.png
+```
+
+> Using the image dimensions are optional, but the [Google Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/) will consider the JSON-LD to have errors when an image does not have specified dimensions.
```
benbalter: This shouldn't be a block quote.

README.md
```diff
@@ -193,19 +194,31 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
...
 ```yml
 image:
+  path: /img/banner.png
   twitter: /img/twitter.png
   facebook: /img/facebook.png
   height: 100
   width: 100
 ```
 
+Or if you have no site specific images simply:
```
benbalter: The use of "site" here is confusing. Perhaps "service", "network" or explicitly call out Facebook and Twitter?

I'd also have the simpler and more common use case first and let users opt into additional complexity as they need it, not the other way around.

lib/template.html
benbalter: For a future iteration, it'd be great if we could calculate the height and width for local images (assuming we can do so safely without adding something like imagemagic).

lib/template.html
```diff
@@ -219,28 +219,60 @@
...
 {% endif %}
 
 {% if page.date %}
     "datePublished": {{ page.date | date_to_xmlschema | jsonify }},
 {% endif %}
 
+{% if page.date_modified %}
+    "dateModified": {{ page.date_modified | date_to_xmlschema | jsonify }},
+{% elsif page.date %}
+    "dateModified": {{ page.date | date_to_xmlschema | jsonify }},
```
benbalter: Why include the date modified if it's really the publish date? Is there value in doing this?

benbalter: Also, I'd suggest checking what field jekyll-last-modified-at uses, and supporting that, if installed and working.

aav7fl: Didn't know about that plugin, I'll add that in.

README.md
```diff
@@ -194,30 +195,33 @@ For most users, setting `image: [path-to-image]` on a per-page basis should be e
...
-  path: /img/banner.png
+  default:
+    path: /img/banner.png
+    height: 100
+    width: 100
+  facebook:
+    path: /img/facebook.png
+    height: 90
+    width: 90
+  twitter: /img/twitter.png
```
aav7fl: I spent most of the day thinking about this and how to handle this, but I feel that I need to propose a change on how images are listed.

The current configuration (below) has problems.

```yml
image:
  twitter: /img/twitter.png
  facebook: /img/facebook.png
  height: 100
  width: 100
```

* It is not clear from the configuration what the height and width are for (The documentation says that the height and width are set for the Facebook image, but we can't infer from the config.)
* If we want to add another image that requires dimensions (i.e. a non-network specific image into the JSON-LD), we must cobble together either a different named field for height and width or create nested fields. This is because there is no guarantee that the two images would have the same dimensions. It would actually make more sense if they had different dimensions, hence the need for a network specific image.

My proposal is: 

```yml
image:
  default:
    path: /img/banner.png
    height: 100
    width: 100
  facebook:
    path: /img/facebook.png
    height: 90
    width: 90
  twitter: /img/twitter.png
```

The way I have it set up, the following will still work:

```yml
image:
  default:
    path: /img/banner.png
  facebook:
    path: /img/facebook.png
  twitter: /img/twitter.png
```

and this will also work.

```yml
image:
  default: /img/banner.png
  facebook:/img/facebook.png
  twitter: /img/twitter.png
```

and this:

```yml
image: /img/banner.png
```

The downside is that **this will break current implementation** of those who use the `height` and `width` fields in the original for Facebook images. So I understand if you reject this idea. 

Unless we decide to use some type of ugly handling... 

Thoughts?

aav7fl:
I would like to propose a different method of handling the images. 
[Conversation starts here](https://github.com/jekyll/jekyll-seo-tag/pull/151#pullrequestreview-22336028)

spec/jekyll_seo_tag_spec.rb
```diff
@@ -141,7 +141,7 @@
 
     context "with page.image as an object" do
       context "when given a path" do
-        let(:page) { make_page("image" => { "path" => "/img/foo.png" }) }
+        let(:page) { make_page("image" => { "default" => "/img/foo.png" }) }
```
aav7fl: I will look into why I changed this test. I shouldn't have done that.

aav7fl:
I spent most of the day thinking about this and how to handle this, but I feel that I need to propose a change on how images are listed.

The current configuration (below) has problems.

```yml
image:
  twitter: /img/twitter.png
  facebook: /img/facebook.png
  height: 100
  width: 100
```

* It is not clear from the configuration what the height and width are for (The documentation says that the height and width are set for the Facebook image, but we can't infer from the config.)
* If we want to add another image that requires dimensions (i.e. a non-network specific image into the JSON-LD), we must cobble together either a different named field for height and width or create nested fields. This is because there is no guarantee that the two images would have the same dimensions. It would actually make more sense if they had different dimensions, hence the need for a network specific image.

My proposal is: 

```yml
image:
  default:
    path: /img/banner.png
    height: 100
    width: 100
  facebook:
    path: /img/facebook.png
    height: 90
    width: 90
  twitter: /img/twitter.png
```

The way I have it set up, the following will still work:

```yml
image:
  default:
    path: /img/banner.png
  facebook:
    path: /img/facebook.png
  twitter: /img/twitter.png
```

and this will also work.

```yml
image:
  default: /img/banner.png
  facebook:/img/facebook.png
  twitter: /img/twitter.png
```

and this:

```yml
image: /img/banner.png
```

The downside is that **this will break current implementation** of those who use the `height` and `width` fields in the original for Facebook images. So I understand if you reject this idea. 

Unless we decide to use some type of ugly handling... 

Thoughts?

aav7fl:
Before Ben or Pat take a look at it, @jhabdas What are your thoughts on the change above? I don't want to fall prey to any scope creep or undesired implementations.

lib/template.html
```diff
@@ -98,11 +102,27 @@
 {% endif %}
 
 {% if page.image %}
-  {% assign seo_page_image = page.image.path | default: page.image.facebook | default: page.image %}
-  {% unless seo_page_image contains "://" %}
-    {% assign seo_page_image = seo_page_image | absolute_url %}
+  {% assign seo_page_image_default = page.image.default.path | default: page.image.default | default: page.image.path | default: page.image %}
```
jhabdas: I'd suggest creating some branch logic around this rather than stacking defaults to ease maintainability and ease future refactoring.

pathawks: I don't mind stacked defaults. It let's me see at a glance what the precedence ordering is :+1:

lib/template.html
```diff
@@ -131,18 +151,27 @@
...
-<meta property="og:image" content="{{ seo_page_image }}" />
-  {% if page.image.height %}
-    <meta property="og:image:height" content="{{ page.image.height }}" />
+{% if page.image %}
+  {% if page.image.facebook %}
+    {% assign local_seo_page_image = seo_page_image_facebook %}
+    {% assign local_seo_page_image_height = seo_page_image_facebook_height %}
+    {% assign local_seo_page_image_width = seo_page_image_facebook_width %}
+  {% else %}
+    {% assign local_seo_page_image = seo_page_image_default %}
```
jhabdas: I'm a little confused what `local_` means here. Is it a temp variable, or does it represent an image set in content front-matter (versus site config)? In either case, it feels funky to assign a _default_ to something that says `local_` sitting under branch logic which says `facebook`. Make sense?

jhabdas:
@aav7fl Great job maintaining backwards compatibility. Getting closer to the finish line. I left a note about branch logic, which I feel if, when refactored, will help the logic flow more smoothly (even if it results in more LOC). Sorry my feedback isn't more actionable. I haven't had time to catch up on the conversations you had with @benbalter and don't want to undo any decisions made already.

lib/template.html
aav7fl: `local_` is used in this case to assign one of the images (and create a fallback in case there is no facebook image) inside of `{% if page.image %}`. It's only used in lines 164 - 169. I pull out the image variable here because if we only checked (for example) the facebook image, but didn't add dimension (and did add dimensions to the `default` image) we would have a facebook image assigned dimension tags of the default image. So here I make sure that each image type is properly assigned its image values from the front matter, and no more. 

Having it before the *meta tag logic* makes sure that one can read the logic below without losing themselves in multiple *if statements* each time. 

Maybe using something other than the name `local_` might help, but I can't think of something that wouldn't load the variable name too much..

jhabdas: Good points, all. I'm a :shipit: if we have the tests updated. Don't want to beat a dead horse.

aav7fl: I'll verify tests this afternoon. Make sure I've got coverage over my code.

aav7fl:
I'm going to check my code coverage this afternoon for tests.

aav7fl:
Alright. I've completed my test coverage. There are some areas that have a bit of overlap, but it makes it clearer when the tests are being run rather than separating my JSON and Meta tag image tests. (Thanks for the suggestion to refactor that, @jhabdas). 

I'm all set.

spec/jekyll_seo_tag_spec.rb
```diff
@@ -158,6 +199,67 @@
...
+        end
+
+        it "outputs the default image JSON object with dimensions" do
+          expect(json_data["image"]["url"]).to eql("http://example.invalid/img/default.png")
+          expect(json_data["image"]["height"]).to eql(3)
+          expect(json_data["image"]["width"]).to eql(4)
+        end
+      end
+
+      # Making sure the facebook image does not inherit the default image dimensions
```
jhabdas: Negative test assertions. Nice guardrail

spec/jekyll_seo_tag_spec.rb
```diff
@@ -189,11 +300,49 @@
...
+      let(:site) { make_site("url" => "http://example.invalid") }
+      let(:page) { make_post("seo" => { "type" => "BlogPosting" }, "permalink" => "/foo/") }
+
+      it "outputs the mainEntityOfPage" do
+        expect(json_data["mainEntityOfPage"]["@type"]).to eql("WebPage")
+        expect(json_data["mainEntityOfPage"]["@id"]).to eql("http://example.invalid/foo/")
+      end
+    end
+
+    context "with seo type is CreativeWork" do
```
jhabdas: Thanks for sneaking this in here

jhabdas:
> Only thing that I'm curious about is if we should warn users trying to fix old and new styles

fix -> mix

aav7fl:
That's a good point. I'll look to see if I can figure out how to present new style change in the documentation while mentioning the old method.

aav7fl:
How about that? I added a code deprecation note in the doc. At first,  I thought that might've been going a little too far, but then I realized completely removing the code example would hurt others reading through the documentation trying to understand how their current code even works when it isn't even mentioned.

README.md
```diff
@@ -185,6 +185,8 @@ The following options can be set for any particular page. While the default opti
   * `name` - If the name of the thing that the page represents is different from the page title. (i.e.: "Frank's Café" vs "Welcome to Frank's Café")
   * `type` - The type of things that the page represents. This must be a [Schema.org type](http://schema.org/docs/schemas.html), and will probably usually be something like [`BlogPosting`](http://schema.org/BlogPosting), [`NewsArticle`](http://schema.org/NewsArticle), [`Person`](http://schema.org/Person), [`Organization`](http://schema.org/Organization), etc.
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
+  * `date_modified` - (YYYY-MM-DD hh:mm) A manual override for the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` output. This is useful when the file timestamp does not match the true time that the content was modified.
+    * A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
```
jhabdas: I tried out `last-modified-at` yesterday and it works quite well (and is performant on a 200 page site), and I think it would be something nice to pick up for the GH pages support eventually, so I like this addition to the docs calling it out.

pathawks: FYI: I don't think GH Pages will ever be able to support *last-modified-at*. https://github.com/github/pages-gem/pull/119

Still a great plugin though, and useful for _us_ to support. :+1:

jhabdas: For continuity, I'm leaving a note here. I believe there may still be a shot for the LMA gem.

jhabdas:
@aav7fl you're not going to like this, but after absorbing [a post](http://ben.balter.com/2016/07/21/removing-a-feature-is-a-feature/) by @benbalter I feel the best course of action to get this PR moving again is to, in a separate PR, remove the ability to set customize Facebook and Twitter images, and leverage the image dimensions so they're not specific to Facebook.

If @benbalter and @pathawks are in agreement, I'll produce a PR to pull out the additional functionality, which this PR can be rebased against and everything simplified without changing the API or introducing any additional complexity.

aav7fl:
But wouldn't that remove functionality from websites already in production using those features?

jhabdas:
@aav7fl not according to [this search](https://github.com/search?utf8=%E2%9C%93&q=image+twitter+facebook+extension%3Amarkdown&type=Code&ref=searchresults) (non-exhaustive). Either way, we'd just bump the major if it was necessary. Based on #77, where the the functionality was introduced, it was intended to help display dimensions across social media networks. But I'm not sold on it personally. Facebook and Twitter (and many others), do their own image massaging anyway as they want their OG images to look pretty -- it's in their best interest.

aav7fl:
I'm fine without the twitter and Facebook specific images. If we removed them, it would have no impact on me. I'm just not sure how @benbalter and @pathawks wish to handle this.

aav7fl:
Bump?

benbalter:
I'm having trouble following what's going on here. It sounds like there are three things being proposed:

1. Adding the author field
2. Adding a date modified field
3. Removing platform-specific dimensions

If that's the case, and unless there's a hidden interdependency, I believe those goals would be best served by implementing the logic as three separate pull requests, where the merits of each could be discussed individually, with being blocked (or confusing) one another.

jhabdas:
@benbalter @aav7fl Here's what I'm suggesting: https://github.com/jekyll/jekyll-seo-tag/pull/174.

aav7fl:
@jhabdas I like that idea. It would help me remove the blocking code and get this accepted if it's ok that there are two features here (Date modified and author field).

aav7fl:
@jhabdas If you get PR #174 into the master branch, I'll begin removing the image portions of my code. (Seem like it would be easier to start from your point that move this backwards)

aav7fl:
@benbalter @jhabdas Removed image changes from this PR. #174 is handling image changes. Should be more straight forward.

All additions are necessary to appease Google Structured Data Testing Tool.

README.md
```diff
@@ -78,6 +78,8 @@ The SEO tag will respect any of the following if included in your site's `_confi
 * `social` - For [specifying social profiles](https://developers.google.com/structured-data/customize/social-profiles). The following properties are available:
   * `name` - If the user or organization name differs from the site's name
   * `links` - An array of links to social media profiles.
+  * `date_modified` - (YYYY-MM-DD hh:mm) A manual override for the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` output. This is useful when the file timestamp does not match the true time that the content was modified.
```
jhabdas: For clarity, I feel we should mention what's being overridden (the `date`?). Also, given Jekyll supports use of ISO 8601 dates, perhaps it's best if we left any formatting out of the example so as not to lead implementations (which might otherwise choose to include the Timezone offset). Also, and now I'm being persnickety, it would be great if this and the next bullet could be condensed into one.

aav7fl: On it.

spec/jekyll_seo_tag_spec.rb
```diff
@@ -237,11 +237,49 @@
     end
 
     context "with absolute site.logo" do
-      let(:site) { make_site("logo" => "http://cdn.example.invalid/logo.png", "url" => "http://example.invalid") }
+      let(:site) { make_site("logo" => "http://cdn.example.invalid/logo.png", "url" => "http://example.invalid", "author" => "Mr. Foo") }
```
jhabdas: Feels like this should be broken into a different context so we're not conflating test logic between existing and new features. Does that make sense?

jhabdas:
This is looking great! Much easier to grok now. Just one small question/comment about the test logic and I'm a :shipit: personally.

spec/jekyll_seo_tag_spec.rb
aav7fl: I guess. But it feels like there are unnecessary LOC when the actual test is given with context in line 246.

jhabdas: I'll defer this one. I don't have the experience with tests to make a good judgement here. Just wafting around code smell. 💨

aav7fl: Maybe we'll save it for when we refactor the tests to bring our coverage back up.

README.md
aav7fl: How about this:

 * `date_modified` - Manually specify the `dateModified` field in the JSON-LD output. This field will take **first priority** for the `dateModified` JSON-LD output. This is useful when the file timestamp does not match the true time that the content was modified. A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.

spec/jekyll_seo_tag_spec.rb
pathawks: I don't think I understand the connection between `author` and `logo`.
If we're adding something new, we should add new tests, _not_ modify existing tests.

spec/jekyll_seo_tag_spec.rb
```diff
@@ -237,11 +237,49 @@
...
         expect(json_data["publisher"]["logo"]["url"]).to eql("http://cdn.example.invalid/logo.png")
       end
+
+      it "outputs the author" do
+        expect(json_data["publisher"]["name"]).to eql("Mr. Foo")
+      end
+    end
+
+    context "with page author" do
+      let(:site) { make_site("logo" => "/logo.png", "url" => "http://example.invalid") }
```
pathawks: Is this necessary here? Context is `"with page author"` says nothing about logo.

lib/template.html
```diff
@@ -65,6 +65,10 @@
   {% assign seo_author_twitter = seo_author_twitter | replace:"@","" %}
 {% endif %}
 
+{% if page.date_modified or page.last_modified_at or page.date %}
+  {% assign seo_date_modified = page.seo.date_modified | default: page.last_modified_at | default: page.date %}
```
pathawks: I don't know if we want to set `seo_date_modified` at all if all we have is `page.date`. What is the motivation?

aav7fl:
Alright. It should be ready now.

lib/template.html
aav7fl: Google will complain if there is no "PageModified", even if it is the exact same. 

> The dateModified field is recommended. Please provide a value if available.

spec/jekyll_seo_tag_spec.rb
aav7fl: Sure. The lovely Google Structured Data Testing Tool complains when the publisher object (required to pass..) does not have a name. For the purposes of a blog (What Jekyll is), the name is the author.

http://i.imgur.com/xnu0JG7.png

pathawks: Sorry, I was referring to the `logo` bits. Is that necessary here?

lib/template.html
pathawks: It would seem that if `date_modified` or similar is not set, we do not have a value available.

aav7fl: I would hope so, but the warning from their tool gives no indication if missing this field is a 'include it if you use it, ignore it if you don't", I can leave it out. But I wanted to remove all warnings in case it had an impact.

aav7fl: Addition: When I check out websites that already use JSON-LD, I note that they use dateModified even if it is the same as the publish date. 

> view-source:http://www.theverge.com/2017/4/3/15123170/bmw-520i-2017-5-series-review

spec/jekyll_seo_tag_spec.rb
aav7fl: Kind of. I *think* I understand what you are asking;

If I have the publisher object
```
"publisher": {"@type": "Organization",
"name": "Kyle Niewiada",
"logo": {"@type": "ImageObject",
"url": "https://www.kyleniewiada.org/assets/img/logo.png"}},
```

I need to have both logo and author together. If I remove just the author, it's an error. If I remove just the logo, it's an error. So they need to be both implemented together. That means when I make the site in the test, I need the logo and the author together.

jhabdas:
beardance niiiiiiice work @aav7fl

aav7fl:
Thanks for working through me with this!

lib/template.html
pathawks: The difference between this plugin and The Verge is that they control every aspect of their site, and this plugin has to serve a more general use case. We have no way of knowing that a document’s date is the same as (or even a good approximation of) its updated date. I assert that the absence of data is better than made up daya.

When you say “Google will complain,” do you mean validation fails, or Google complains loudly? If validation fails, I guess we'll do what's needed to fix that. If it's merely a warning, that seems like A Good Thing! Publishers _should_ be made aware that including modified information is an important thing to do, and we can't just hand-wave over that.

aav7fl: It's merely a warning. 

That is a valid point that the document's date is not the same as its upload date. 

I will pull that last default snippet to exclude the `page.date`.

aav7fl:
@pathawks Date handling complete. Thanks for pointing out that the file date may not properly represent the publish date; and all of the other valid points that I missed along the way!

benbalter:
Think this can be merged now?

aav7fl:
It looks good to me. 😃

pathawks:
@benbalter Yes, LGTM. Sorry; I didn't realize that I was the one holding things up.

@aav7fl Thanks for making those changes, and for your patience. This will be a great addition to the plugin :+1::tada:

benbalter:
Thanks @aav7fl! 🥇 🌵 🍰

benbalter:
Going to get a new release out right now with this in it.

jhabdas:
\o/

PR #153, opened 2017-01-17
Consolidate absolute URL into one assignment

### LINKED ISSUE #152: Consolidate `page.url` in template
Several places in the template call `{{ page.url | replace:'/index.html','/' | absolute_url }}`

Perhaps we could consolidate all of these into one assignment to `seo_page_url` so the filters are only called once instead of each time. This would DRY up the template semantically and may make generation a _tiny_ bit faster.

pathawks:
Fixes #152

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -1,5 +1,7 @@
 <!-- Begin Jekyll SEO tag v{{ seo_tag.version }} -->
 
+{% assign seo_page_url = page.url | replace:'/index.html','/' | absolute_url %}
+
 {% if page.url == "/" or page.url == "/about/" %}
   {% assign seo_homepage_or_about = true %}
 {% endif %}
@@ -110,8 +112,8 @@
 {% endif %}
 
 {% if page.url %}
-  <link rel="canonical" href="{{ page.url | replace:'/index.html','/' | absolute_url }}" />
-  <meta property="og:url" content="{{ page.url | replace:'/index.html','/' | absolute_url }}" />
+  <link rel="canonical" href="{{ seo_page_url }}" />
+  <meta property="og:url" content="{{ seo_page_url }}" />
 {% endif %}
 
 {% if seo_site_title %}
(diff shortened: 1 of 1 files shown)


lib/template.html
```diff
@@ -110,8 +112,8 @@
 {% endif %}
 
 {% if page.url %}
```
pathawks: When will a page not have a URL?

lib/template.html
```diff
@@ -1,5 +1,7 @@
 <!-- Begin Jekyll SEO tag v{{ seo_tag.version }} -->
 
+{% assign seo_page_url = page.url | replace:'/index.html','/' | absolute_url %}
+
 {% if page.url == "/" or page.url == "/about/" %}
```
pathawks: Should we ever check for `/index.html` here?

lib/template.html
DirtyF: Is there a case when we don't wan't to add `canonical` or `og:url` ? If not 👍  we can get rid of this condition in a separate PR.

lib/template.html
DirtyF: I guess it's because the default `page.url` for a Jekyll site `index.html` is `/index.html`

lib/template.html
benbalter: Agree it's incorrect as written. If `site.url` isn't set, the output may be a little silly. The relevant commit is https://github.com/jekyll/jekyll-seo-tag/commit/df09740757543e6f5a5cf59401c9c7ec7e67362f.

I believe that should be `site.url`, not `page.url`, which means we'd actually have a canonical URL.

lib/template.html
benbalter: I *think* we've run into this with other plugins. This is an edge case when the use sets the permalink to `/index.html` via YAML front matter. We should check for it.

benbalter:
Although a great implementation, I don't believe this is necessary in a post-drop world.

Issue #161, opened 2017-02-25
Multi-author support


jhabdas:
Thoughts potential template and test refactor to ease future updates after #151 and prep for enhanced support for multi-author sites like minimal mistakes? Some initial observations:

- Looks like authors.yml only supports Twitter authors and not Facebook/OG settings
- The og:publisher is present but not the og:author, and it's not clear how to use publisher on multi-author sites (at least to me)
- OG variables in templates are carrying Facebook names, which seems not right
- A cleanse may help uncover some easy wins WRT gaps in existing functionality

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

jhabdas:
In a future reality

PR #162, opened 2017-03-01
Update title separator. Closes #121

### LINKED ISSUE #121: Allow customization of the title-separator
Hey guys
Currently the `{{ page.title }}`-Property and the `{{ site.title }}`-Property inside the `<title>`-Tag are allways separated with an `-`-Character.
Could you implement a way to customize the separator to use?
I'd, for example, like to use an `·` instead of an `-`

Best regards an' thanks in advance

jhabdas:
Replaces the "-" used as a title separator with "|".
Ref: https://moz.com/learn/seo/title-tag

Also updates docs to <s>specify title optional and fix errata</s> provide additional info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A Jekyll plugin to add metadata tags for search engines and social networks to b
 
 Jekyll SEO Tag adds the following meta tags to your site:
 
-* Pages title (with site title appended when available)
+* Page title, with site title or description appended
 * Page description
 * Canonical URL
 * Next and previous URLs on paginated pages
--- a/lib/template.html
+++ b/lib/template.html
@@ -10,7 +10,7 @@
 
 {% if page.title and seo_site_title %}
   {% assign seo_title = seo_title | append:" | " | append: seo_site_title %}
-{% elsif site.description %}
+{% elsif site.description and seo_site_title %}
   {% assign seo_title = seo_title | append:" | " | append: site.description %}
 {% endif %}
 
--- a/lib/template.html
+++ b/lib/template.html
@@ -9,9 +9,9 @@
 {% assign seo_title = page.title | default: seo_site_title %}
 
 {% if page.title and seo_site_title %}
-  {% assign seo_title = seo_title | append:" | " | append: seo_site_title %}
+  {% assign seo_title = page.title | append:" | " | append: seo_site_title %}
 {% elsif site.description and seo_site_title %}
-  {% assign seo_title = seo_title | append:" | " | append: site.description %}
+  {% assign seo_title = seo_site_title | append:" | " | append: site.description %}
 {% endif %}
 
 {% if page.seo and page.seo.name %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -5,21 +5,13 @@
 {% endif %}
 
 {% assign seo_site_title = site.title | default: site.name %}
+{% assign seo_page_title = page.title | default: seo_site_title %}
+{% assign seo_title = page.title | default: seo_site_title %}
 
-{% if page.title %}
-  {% assign seo_title = page.title %}
-  {% assign seo_page_title = page.title %}
-
-  {% if seo_site_title %}
-    {% assign seo_title = seo_title | append:" - " | append: seo_site_title %}
-  {% endif %}
-{% elsif seo_site_title %}
-  {% assign seo_title = seo_site_title %}
-  {% assign seo_page_title = seo_site_title %}
-
-  {% if site.description %}
-    {% assign seo_title = seo_title | append:" - " | append: site.description %}
-  {% endif %}
+{% if page.title and seo_site_title %}
+  {% assign seo_title = seo_title | append:" | " | append: seo_site_title %}
+{% elsif site.description %}
+  {% assign seo_title = seo_title | append:" | " | append: site.description %}
 {% endif %}
 
 {% if page.seo and page.seo.name %}
--- a/spec/jekyll_seo_tag_spec.rb
+++ b/spec/jekyll_seo_tag_spec.rb
@@ -44,13 +44,47 @@ describe Jekyll::SeoTag do
       expect(output).to match(expected)
     end
 
+    context "with site.name" do
+      let(:site) { make_site("name" => "Site Name") }
+
+      it "builds the title with a page title and site name" do
+        expect(output).to match(%r!<title>foo \| Site Name</title>!)
+      end
+    end
+
     context "with site.title" do
       let(:site) { make_site("title" => "bar") }
 
       it "builds the title with a page title and site title" do
-        expect(output).to match(%r!<title>foo - bar</title>!)
+        expect(output).to match(%r!<title>foo \| bar</title>!)
       end
     end
+
+    context "with site.title and site.description" do
+      let(:site) { make_site("title" => "Site Title", "description" => "Site Description") }
+
+      it "builds the title with a page title and site title" do
+        expect(output).to match(%r!<title>foo \| Site Title</title>!)
+      end
+
+      it "does not build the title with the site description" do
+        expect(output).not_to match(%r!<title>foo \| Site Description</title>!)
+      end
+    end
+
+    context "with site.title and site.description" do
+      let(:site) { make_site("title" => "Site Title", "description" => "Site Description") }
+
+      it "builds the title with a page title and site title" do
+        expect(output).to match(%r!<title>foo \| Site Title</title>!)
+      end
+
+      it "does not build the title with the site description" do
+        expect(output).not_to match(%r!<title>Page Title \| Site Description</title>!)
+      end
+
+    end
+
   end
 
   context "with site.title" do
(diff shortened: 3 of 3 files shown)


jhabdas:
Will get those style snafus cleaned up as I incorporate any feedback on the Pull. Thank you kindly.

lib/template.html
```diff
@@ -5,21 +5,13 @@
...
-  {% if seo_site_title %}
-    {% assign seo_title = seo_title | append:" - " | append: seo_site_title %}
-  {% endif %}
-{% elsif seo_site_title %}
-  {% assign seo_title = seo_site_title %}
-  {% assign seo_page_title = seo_site_title %}
-
-  {% if site.description %}
-    {% assign seo_title = seo_title | append:" - " | append: site.description %}
-  {% endif %}
```
benbalter: Can you explain what's going on here?

README.md
```diff
@@ -6,10 +6,10 @@ A Jekyll plugin to add metadata tags for search engines and social networks to b
 
 ## What it does
 
-Jekyll SEO Tag adds the following meta tags to your site:
+Jekyll SEO Tag adds the following tags to your site:
 
-* Pages title (with site title appended when available)
-* Page description
+* Optional `title` delimited with a pipe
+* Meta description
```
pathawks: There are some things going on here that are outside the scope of this PR.

Meta in this sense is only meaningful to HTML developers. I suspect “Page Description” is more clear to most users.

spec/jekyll_seo_tag_spec.rb
```diff
@@ -36,21 +36,43 @@
   end
 
   context "with page.title" do
-    let(:page) { make_page("title" => "foo") }
+    let(:page) { make_page("title" => "Page Title") }
```
pathawks: I'm not a fan of changing behavior and tests together in the same commit.

lib/template.html
jhabdas: Yes, I refactored out 8 lines of code, and covered them all with tests.

README.md
jhabdas: SGTM. Will update and remove jargon.

spec/jekyll_seo_tag_spec.rb
jhabdas: Will pull from PR

jhabdas:
Review feedback incorporated, offenses resolved.

lib/template.html
```diff
@@ -5,21 +5,13 @@
...
+{% if page.title and seo_site_title %}
+  {% assign seo_title = seo_title | append:" | " | append: seo_site_title %}
+{% elsif site.description %}
+  {% assign seo_title = seo_title | append:" | " | append: site.description %}
```
pathawks: I don't think this does the same thing as what we originally intended.

The goal was for the title to be `page.title -site.title` or `site.title - site.description`

The only time `site.description` is in the title is when `page.title` is empty.

If `page.title` is not empty and `site.title` is empty, the title should just be `page.title`

Right?

jhabdas: I added a negative test assertion on L70 to ensure this was always the case, dropping the AND check here as it wasn't doing anything after the refactor. This was done as a trade-off to improve readability over the previous branch logic removed while slightly obfuscating functionality now covered by tests.

pathawks: That does not appear to be what L70 is testing.

I am talking about the case where `page.title` is set, `site.name` is not set, `site.description` is set.

In this case, the first `if` condition is false because there is no `seo_site_title`, but the `elsif` is true because `site.description` does exist.

So our title becomes `page.title | site.description` :-1:

jhabdas: Good eye, @pathawks. Resolved with https://github.com/jekyll/jekyll-seo-tag/pull/162/commits/51db5806c20f3947936a009afca8ae4a6c6230c2

lib/template.html
```diff
@@ -5,21 +5,13 @@
...
-  {% assign seo_title = seo_site_title %}
-  {% assign seo_page_title = seo_site_title %}
-
-  {% if site.description %}
-    {% assign seo_title = seo_title | append:" - " | append: site.description %}
-  {% endif %}
+{% if page.title and seo_site_title %}
+  {% assign seo_title = seo_title | append:" | " | append: seo_site_title %}
+{% elsif site.description and seo_site_title %}
+  {% assign seo_title = seo_title | append:" | " | append: site.description %}
```
pathawks: Can we use `seo_site_title` here, just to make explicit what we are doing?

```
{% assign seo_title = seo_site_title | append:" | " | append: site.description %}
```

We could probably get rid of the assignment on L9 and also change l12 to refer to `page.title`

Since it is what we're testing for, it seems like that should be what we use.

jhabdas: Agreed. Updated with https://github.com/jekyll/jekyll-seo-tag/pull/162/commits/2a48f272f68b309b108b4cda27a862e6744c1f20. Had to keep L9 as it contains some default logic which, if would remove, would break some of the tests introduced.

pathawks:
I'm still pretty indifferent on the specific title separator, but the implementation looks good to me.

@benbalter?

benbalter:
@jekyllbot: merge +minor.

jhabdas:
My pleasure, and thanks for letting the refactoring in with the PR. It's pragmatic and helps address the refactoring I wanted to tackle for #161. I ran the updated test suite against the previous template logic after a replacing all instances of `\|` with `-` in the spec and we're 1:1 with previous functionality, now with better codecov.

benbalter:
@jhabdas looks like it didn't actually merge. Can you undelete the branch so I can reopen the PR and merge manually?

jhabdas:
@benbalter done

nhaglind:
What's the deal with the change here? I have a number of sites that use `|` as a title separator `About Something | Company Name` and now after the latest update that has been removed in serving and the build and becomes `About Something Company Name`. Even if it's specifically included in front matter.

benbalter:
> Even if it's specifically included in front matter.

@nhaglind you put both the page title and the site title in the front matter? Can you provide a link or code snippet to how you're using this? I don't see how it would affect that behavior.

jhabdas:
@nhaglind titles are optional with this plugin. If you were overriding to achieve the pipe separator you may already be customized. Please check release notes and README for usage, and file a bug if you think you've found an edge case.

nhaglind:
Yes, absolutely. 

```
---
permalink: /foobar-noodles
title: Foobar Noodles | ACME Corp
description: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce sit amet dui diam.
layout: default
image: /images/noodles.jpg
tags: ['foo', 'bar',]
---
```

This previously returned the title `Foobar Noodles | ACME Corp` and now returns `Foobar Noodles ACME Corp`.

nhaglind:
A simple example of the behavior that should be reproducible using default of `{% seo %}`. 

`title: Foobar Noodles | ACME Corp` = `<title>Foobar Noodles ACME Corp</title>`
`title: Foobar Noodles - ACME Corp` = `<title>Foobar Noodles - ACME Corp</title>`

benbalter:
@nhaglind does this happen in 2.2.0 or in 2.2.1 (just released moments ago) or both? Also, what Markdown engine are you using (I'm wondering if it's treating it as markdown).

nhaglind:
@benbalter using latest version of jekyll 3.4.3 with default kramdown. This is related to 2.2.0 I'll update to 2.2.1 right now to see if this changes behavior.

benbalter:
I was able to recreate this locally. The issue is that `"Foo | Bar" | markdownify`, at least with Kramdown, results in `Foo Bar`. While it's not the plugin's intention for users to put both the page title and site title in the page's `title` field, you can get around the issue by escaping the pipe, e.g., `Foo \| Bar`, which testing locally, should work.

nhaglind:
You're correct escaping fixes `title` and `og` but it for Twitter it will include the escaping character. `<meta name="twitter:title" content="Foobar Noodles \| ACME Corp">`.

What would properly format all these things together? I have no problem doing an adjustment like the following but this update resulted in a sitewide change. 

`site: ACME Corp` in `_config.yml` then in each page `title: Foobar Noodles`?

Looking to achieve what was previously being returned prior to update.

benbalter:
> `site: ACME Corp` in `_config.yml` then in each page `title: Foobar Noodles`?

Correct. That's the intended use (except: `title` in `_config.yml` instead of `site`). 

I believe with 2.2.1, they should be consistently escaped (as they use the same `markdownify` filter.

nhaglind:
Looks like I got a lot of pages to update! But thanks a lot @benbalter for taking the time to look into this behavioral change.

jhabdas:
@nhaglind You may want to consider updating your titles now to make use of the `|` syntax afforded by this update (since it seems like that's what you were trying to achieve). Please see the following resource for best practices on page titles: https://moz.com/learn/seo/title-tag

cup:
https://github.com/jekyll/jekyll-seo-tag/issues/121

cup:
https://github.com/jekyll/jekyll-seo-tag/issues/121
https://github.com/jekyll/jekyll-seo-tag/issues/331
https://github.com/jekyll/jekyll-seo-tag/pull/162

PR #164, opened 2017-03-18
Remove html comment in production

### LINKED ISSUE #143: Embedded comment necessary?
Not a breaking issue and there is almost 0 impact to performance, but is there a reason this is needed? Often times visible production comments are only for semantic purposes. 

I'm referring to the following

```
<!-- Begin Jekyll SEO tag v2.1.0 -->
<!-- End Jekyll SEO tag -->
```

mkutyba:
Closes #143

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -1,4 +1,6 @@
-<!-- Begin Jekyll SEO tag v{{ seo_tag.version }} -->
+{% if jekyll.environment != "production" %}
+  <!-- Begin Jekyll SEO tag v{{ seo_tag.version }} -->
+{% endif %}
 
 {% if page.url == "/" or page.url == "/about/" %}
   {% assign seo_homepage_or_about = true %}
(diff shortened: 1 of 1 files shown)


benbalter:
@mkutyba Can you explain a bit more about your motivation here? Why does that motivation outweigh the benefit of having the comment (e.g., debugging, discoverability, etc.)?

mkutyba:
First thing is that it's useless in production environment. Second is that it is still present outside production, so you preserve all benefits of having the comment.

DirtyF:
As Ben says, it's a way to show that SEO tags are generated by `jekyll-seo-tag` plugin. It's here for visibility. That's also what Yoast SEO Wordpress plugin do:

```<!-- This site is optimized with the Yoast SEO Premium plugin v4.5 - https://yoast.com/wordpress/plugins/seo/ -->
```

It might be useless for _you_ but it's primarily intended to promote the use of this great plugin. 
A comment is a small price to pay to benefit from these features, don't you think? 🤑

7php:
hey @mkutyba, just use `jekyll-minifier` for production, problem solved!

DirtyF:
Thanks @7php for pointing out to this solution, closing then.

PR #165, opened 2017-03-31
Added site.logo as an alternate OG image (if page.image doesn’t exist)

aarongustafson:


aarongustafson:
Just saw the tests… fixing now.

aarongustafson:
Looks like Rubocop is barking at master (stuff I didn’t touch). Do you want me to fix the issues?

aarongustafson:
Fixes #168

DirtyF:
@aarongustafson nice contribution ❤️ 

`rubocop -a` should fix some offenses but if this is not related to your PR this can be fixed later in a dedicated PR

pathawks:
> this is not related to your PR this can be fix later in a dedicated PR

:+1:

DirtyF:
@pathawks done in #170

aarongustafson:
@DirtyF Thanks!

benbalter:
See https://github.com/jekyll/jekyll-seo-tag/issues/168#issuecomment-290896438:

> The reason we want the user to explicitly define an image instead of just falling back to the logo or something is because there may be instances where a user would rather include no image than include the logo. If this is not the case, the user can easily specify a site-wide default image to use.
 
Thoughts?

aarongustafson:
@benbalter Somehow I missed that in the docs. The PR is unnecessary.

PR #166, opened 2017-03-31
Add og:locale support

aarongustafson:


aarongustafson:
Addresses #167

benbalter:
Is there a pattern for what the key should be called? Can we use `site.lang` (and `post.lang`) [to share a setting with `jekyll-feed`](https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L5)?

aarongustafson:
@benbalter It is not a standard config in Jekyll (it was in Octopress), but I’m absolutely game to piggyback on `jekyll-feed`. PR update forthcoming.

aarongustafson:
@benbalter Made the changes & merged back your `master` branch.

.gitignore
```diff
@@ -7,4 +7,5 @@
 /pkg/
 /spec/reports/
 /tmp/
+/bin/
```
pathawks: What is this?

README.md
```diff
@@ -102,12 +102,15 @@ webmaster_verifications:
...
 
+* `locale` - The locale these tags are marked up in. Of the format `language_TERRITORY`. Default is `en_US`.
+
 The SEO tag will respect the following YAML front matter if included in a post, page, or document:
 
 * `title` - The title of the post, page, or document
 * `description` - A short description of the page's content
 * `image` - URL to an image associated with the post, page, or document (e.g., `/assets/page-pic.jpg`)
 * `author` - Page-, post-, or document-specific author information (see below)
+* `locale` - Page-, post-, or document-specific language information
```
benbalter: This will need to be `lang`. We should also indicate that this can be `site.lang`.

pathawks: This should be changed to `lang`, correct?

lib/template.html
```diff
@@ -97,6 +97,8 @@
   {% assign seo_page_image = seo_page_image | escape %}
 {% endif %}
 
+{% assign seo_page_locale = page.lang | default: site.lang | default: "en_US"" %}
```
pathawks: Is there a reason we need to default to `"en_US"` instead of just not setting anything if the user didn't ask for it?

pathawks: Can we instead call this `seo_page_lang` for consistency?

.gitignore
aarongustafson: I was following [the RSpec install recommendation](http://rspec.info/) which created that directory. I didn’t want to bundle it with the repo.

README.md
aarongustafson: Yes. Fix coming. Sorry I missed that.

aarongustafson: Yep.

lib/template.html
aarongustafson: Yes we can.

The [OG docs](https://developers.facebook.com/docs/reference/opengraph/object-type/website/) default to “en_US”. I thought it made sense to be explicit. If you feel otherwise, we can drop the tag without a defined language (or even if it is “en_US”).

lib/template.html
```diff
@@ -97,6 +97,8 @@
   {% assign seo_page_image = seo_page_image | escape %}
 {% endif %}
 
+{% assign seo_page_lang = page.lang | default: site.lang | default: "en_US" %}
```
pathawks: I don't know about default values here, but I suppose I don't have a strong preference either way.

mmistakes: Should there be some sort of string normalization applied with maybe a `| replace: '-', '_'`?

For example it's possible to have a `site.lang` of `en-US` for the `lang` attribute in `<html lang="...">`. Same for `xml:lang` in an Atom feed `<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="...">`.

In both cases it appears they expect values of language-REGION, e.g. `en-US`, where as `og:locale` wants `en_US` with an underscore.

- https://www.w3.org/International/articles/language-tags/
- https://developers.facebook.com/docs/opengraph/guides/internationalization/

aarongustafson: Good call. I can make that happen.

aarongustafson: And done (assuming I didn’t do anything to annoy Rubocop).

benbalter:
Nice work @aarongustafson! 🌐 🎏 🈂️

aarongustafson:
Yay! Any idea when the next release will be?

benbalter:
If https://github.com/jekyll/jekyll-seo-tag/pull/151 lands soon, it'd be a solid release to get that in.

PR #171, opened 2017-03-31
No need to delete Jekyll's version from Gemfile

DirtyF:
🌸  Some spring cleanup as Github Pages now supports Jekyll 3.4.3 🌸 

/cc @benbalter


benbalter:
Thanks @DirtyF! ⚡️ ⚙️ 🐎

PR #172, opened 2017-03-31
Bump dev dependencies

DirtyF:
Q: Any possibility to keep this in sync with Jekyll?

### DIFF
--- a/jekyll-seo-tag.gemspec
+++ b/jekyll-seo-tag.gemspec
@@ -27,9 +27,9 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "jekyll", "~> 3.3"
-  spec.add_development_dependency "bundler", "~> 1.10"
-  spec.add_development_dependency "rake", "~> 10.0"
-  spec.add_development_dependency "rspec", "~> 3.3"
-  spec.add_development_dependency "html-proofer", "~> 2.5"
+  spec.add_development_dependency "bundler", "~> 1.14"
+  spec.add_development_dependency "rake", "~> 12.0"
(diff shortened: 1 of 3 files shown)


DirtyF:
@pathawks 🙁

jekyll-seo-tag.gemspec
```diff
@@ -1,34 +1,35 @@
...
+  spec.require_paths = ["lib"]
 
-  spec.add_dependency 'jekyll', '~> 3.3'
-  spec.add_development_dependency 'bundler', '~> 1.10'
-  spec.add_development_dependency 'rake', '~> 10.0'
-  spec.add_development_dependency 'rspec', '~> 3.3'
-  spec.add_development_dependency 'html-proofer', '~> 2.5'
-  spec.add_development_dependency 'rubocop', '~> 0.37'
+  spec.add_dependency "jekyll", "~> 3.3"
+  spec.add_development_dependency "bundler", "~> 1.14"
```
pathawks: Is there a reason for us to _not_ allow `bundler 1.10`? The pessimistic operator will allow anything up to (but not including) `2.0`, so we only want to increment this if we have a reason to cut off support for a particular version.

DirtyF: 1.14 fixed an annoying bug in bundler https://github.com/bundler/bundler/issues/5154

benbalter:
> Any possibility to keep this in sync with Jekyll?

What do you mean?

DirtyF:
@benbalter When a plugin shares dependencies with jekyll, can it inherit from those? For instance `jekyll-seo-tag` depends on `bundler` like `jekyll`, does rubygems allows to say that `jekyll-seo-tag `and `jekyll` should share the same `bundler` version?

PR #174, opened 2017-04-03
refactor(template): use og:image for twitter

jhabdas:
#77 and #78 added support for different images on Twitter and Facebook, an advanced feature of SEO Tag Gem. The implementation has led to complications updating JSON-LD data in #151. Under Pareto principal (I also did a quick and dirty search on GitHub) we can reasonably assume this feature is not often used, and, as such, should be removed. The result of which simplifies template logic, offers a fallback experience by leveraging Twitter's use of `og:image` in place of `twitter:image` (sees [twitter card tag ref](https://dev.twitter.com/cards/markup)) and helps unblock #151.

Users can no longer specify a separate twitter and Facebook image. If both a Facebook (Open Graph) and Twitter image are provided in the `image` object, the Facebook image will take precedence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ Jekyll SEO Tag adds the following meta tags to your site:
 * Canonical URL
 * Next and previous URLs on paginated pages
 * [JSON-LD Site and post metadata](https://developers.google.com/structured-data/) for richer indexing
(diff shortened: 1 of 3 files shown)


benbalter:
Thanks @jhabdas! 🎉 🌮 🦃

benbalter:
@jekyllbot: merge +minor

PR #182, opened 2017-04-04
Add plugin URL to comment

DirtyF:
Following the discussion in #164, why not show some pride and information?

/cc @benbalter @pathawks


pathawks:
meh.

nhaglind:
I think the current comment is better. It's clear and concise, tells me when it starts and ends. Optimized sounds tacky, and just because yoast does it doesn't mean it's right. Also, the only people concerned with source code would be developers and when they see the current comment they know what to look for.

Issue #176, opened 2017-04-03
Move seo_author_* logic to Ruby


benbalter:
The top 100 or so lines of the template is getting a bit unwieldy and I can't imagine it's performant. We should centralize the logic in ruby for figuring out the author (post, data, or global), and any other super-safe hash or string manipulation.

pathawks:
We had talked previously about starting yet another gem to deal with author logic, as anything we do here would also apply to **jekyll-feed** and likely themes as well.

benbalter:
> We had talked previously about starting yet another gem to deal with author logic

Yes! I was thinking step one would be to move it to Ruby here, and then step two could potentially move that to be shared some way once we know it works (e.g., generally make `post.author` a resolved object available to generators).

PR #184, opened 2017-04-08
Convert seo_author_* logic to a drop

benbalter:
This PR moves the `seo_author_*` logic in the template to Ruby, where it can be implemented as a drop. The goal is readability, testability, and performance.

The resulting output should be identical, save the absence of whitespace in the JSON-LD output.

I believe the original tests should pass with the small whitespace change.

One thing to call out, I'm not 100% sold on the JSON-LD approach, but was trying to find the most efficient way to generate the hash and DRY things up.

Thoughts?

Fixes https://github.com/jekyll/jekyll-seo-tag/issues/176.


DirtyF:
> The resulting output should be identical, save the absence of whitespace in the JSON-LD output.

@benbalter This looks great, I dig the move to Ruby.

I notice minor differences between 2.2.0 and this PR (after beautifying the content in Atom)

diff-jekyll-seo-tag-drop

If I copy the JSON-LD output to https://search.google.com/structured-data/testing-tool I got fewer errors than before. (errors with auhor and pubisher name and type are gone ✅ ) 

 - datePublished is not present
- image.itemtype attribute has an incorrect value

So apart from email that went missing, It's looks rather pretty good to me. 👏 👏 👏

benbalter:
@DirtyF Nice. It looks like what should be the name string is being returned as the entire author hash in the new version? How are you supplying the author? Looking at the code and tests I'm not sure how to recreate that. Unless I'm reading the diff backwards, in which case, shouldn't the `name` field be a string, not a hash with the key `name` in it?

Added `datePublished` back in via d2f053d. Good catch.

benbalter:
Also to note, testing with my own site, I'm consistently seeing a 30%-40% performance improvement over `master`.

DirtyF:
> It looks like what should be the name string is being returned as the entire author hash in the new version? 

Sorry for the confusion, to be clearer the new version returns:

```json 
"author": {"@type": "Person",
"name": {"name":"Frank Taillandier","email":"frank@taillandier.me","twitter":"DirtyF"}},
"description": "Consultant web",
"publisher": {"@type": "Organization",
"name": {"name":"Frank Taillandier","email":"frank@taillandier.me","twitter":"DirtyF"},
```

How are you supplying the author?

Via a key in the config file (setting inherited from the Lanyon theme) https://github.com/poole/lanyon/blob/master/_config.yml#L16-L19

https://github.com/DirtyF/frank.taillandier.me/blob/c0857780c6c8dcda04104d6bb86f939b424ae751/_config.yml#L84-L87

benbalter:
> the new version returns

What is it supposed to return?

spec/jekyll_seo_tag/json_ld_spec.rb
```diff
@@ -0,0 +1,116 @@
+RSpec.describe Jekyll::SeoTag::JSONLD do
+  before do
+    Jekyll.logger.log_level = :error
+  end
+
+  let(:metadata) do
+    {
+      "title"       => "title",
+      "author"      => "author",
```
pathawks: We should add a test for `author` provided as a `Hash` in `metadata`

DirtyF:
@benbalter Everything is fine.

https://search.google.com/structured-data/testing-tool displays one warning left related to image in `BlogPosting` schema but it was already the case before this PR.

lib/jekyll-seo-tag.rb
```diff
@@ -1,7 +1,12 @@
+require "jekyll"
 require "jekyll-seo-tag/version"
 
 module Jekyll
   class SeoTag < Liquid::Tag
+    autoload :JSONLD,  "jekyll-seo-tag/json_ld"
+    autoload :Drop,    "jekyll-seo-tag/drop"
+    autoload :Filters, "jekyll-seo-tag/filters"
```
parkr: I recently ran into an issue with this because I didn't add `lib` to the `$LOAD_PATH`, so local tests didn't work.

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+
+      def version
+        Jekyll::SeoTag::VERSION
+      end
+
+      # Should the `<title>` tag be generated for this page?
+      def title?
+        return false unless title
+        @text !~ %r!title=false!i
```
parkr: This would probably be good to memoize, right?

```ruby
def title?
  return false unless title
  return @is_title if defined?(@is_title)
  @is_title = @test !~ %r!title=false!i
end
```

?

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+
+      # Page title without site title or description appended
+      def page_title
+        @page_title ||= format_string(page["title"] || site_title)
+      end
+
+      # Page title with site title or description appended
+      def title
+        @title ||= begin
+          if page["title"] && site_title
```
parkr: With jekyll-titles-from-headings have populated this by the time this code is called? If so, then 😍!!!

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+          else
+            page_title || site_title
+          end
+        end
+      end
+
+      def name
+        @name ||= begin
+          return seo_name if seo_name
+          return unless homepage_or_about?
```
parkr: Note that none of these `nil` values will be memoized.

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+          if site["social"] && site["social"]["name"]
+            format_string site["social"]["name"]
+          elsif site_title
+            format_string site_title
+          end
+        end
+      end
+
+      def description
+        @description ||= begin
```
parkr: Is the begin here necessary? I believe you can keep this on 2 lines if needed without the `begin...end`

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+      def page
+        @page ||= @context.registers[:page].to_liquid
+      end
+
+      def site
+        @site ||= @context.registers[:site].site_payload["site"].to_liquid
+      end
+
+      def homepage_or_about?
+        ["/", "/index.html", "/about/"].include? page["url"]
```
parkr: are `about.html`, `/index.htm`, `/index/` also valid?

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+
+      attr_reader :context
+
+      def fallback_data
+        {}
```
parkr: Probably good to memoize this

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+
+      def absolute_url?(string)
+        string.include? "://"
```
parkr: @pathawks used `Addressable` for this: https://github.com/jekyll/jekyll/blob/8b8ad61097e449a2176bae031082830adf27590e/lib/jekyll/filters/url_filters.rb#L13

lib/jekyll-seo-tag/drop.rb
```diff
@@ -0,0 +1,239 @@
...
+      end
+
+      def format_string(string)
+        FORMAT_STRING_METHODS.each do |method|
+          string = filters.public_send(method, string)
+        end
```
parkr: This might be a good use-case for `reduce`, or `each_with_object`?

spec/jekyll_seo_tag/json_ld_spec.rb
benbalter: https://github.com/jekyll/jekyll-seo-tag/pull/184/files#diff-b55c10839692a05f48657272f873a3d7R53

benbalter: 791db51

lib/jekyll-seo-tag.rb
benbalter: [Gotcha covered](https://github.com/jekyll/jekyll-seo-tag/blob/master/spec/spec_helper.rb#L1)

lib/jekyll-seo-tag/drop.rb
benbalter: 97% sure? A generator should be be called the render phase.

lib/jekyll-seo-tag/drop.rb
benbalter: ee1b03c

lib/jekyll-seo-tag/drop.rb
benbalter: 0a70e8c

lib/jekyll-seo-tag/drop.rb
benbalter: ab916c9

lib/jekyll-seo-tag/drop.rb
benbalter: Moved to regex and added tests via cfe7e56.

I'm not super excited to support `/index/` or `about.html`. (GitHub Pages lazy indexes and ugly permalinks). I could see a case for the second if there was strong desire.

lib/jekyll-seo-tag/drop.rb
benbalter: Ahh, good call. 8cb5953.

lib/jekyll-seo-tag/drop.rb
benbalter: TIL'd 650fe5c.

lib/jekyll-seo-tag/drop.rb
benbalter: 993eafc

Issue #190, opened 2017-04-11
JSON-LD errors


aav7fl:
In the latest commits (f048342e2296f6a8a915b27eef3326d28e02d7a0), there are a few JSON-LD errors with the Google Structured Data Testing tool when I test against my website. 

1. Author name is missing (Probably because the template is only looking for the `author.name` and not any of the other methods. e.g. `{% assign seo_author = page.author | default: page.authors[0] | default: site.author %}` )
2. Publisher name is missing (same as above)
3. Alt text is now added in the `ImageObject` while the Google Structured Data Testing tool prohibits this. "`The property alt is not recognized by Google for an object of type ImageObject.`'

Screenshot log: http://i.imgur.com/TFvXyn4.png

Let me know if/how you would like to handle those.

benbalter:
Are you able to push up failing tests to reproduce the above? 

For the first, it should be looking to the other means.

aav7fl:
I can try that tomorrow. Will need to freshen up on my Ruby first.

aav7fl:
PR #191 opened with failing test for image tag. 

PR also offers explanation for why the `author` was populated as `null` and why the image object has extra tags.

dkmooers:
I'm having this issue as well. Blog posts are not outputting Author or Publisher at all into JSON-LD, presumably due to OP's suggested reason.

aav7fl:
For me, https://github.com/jekyll/jekyll-seo-tag/pull/192 fixes this (I closed this issue prematurely). Just need to get it pulled into master and the gem updated if it causes no other problems. That should solve your issue, @dkmooers.

aav7fl:
@dkmooers Looks like the PR has been merged and the latest Gem updated with these changes. My issues are solved. Re-open if you're still having an issue.

dkmooers:
Perfect, thanks all!

PR #191, opened 2017-04-12
Failing test for #190

aav7fl:
This PR includes a failing test for issue #190 

This covers:
- JSON alt text

The JSON Alt Text was showing up in my JSON-LD because I have an `alt` tag for my website images inside of the `image` tag for other optimization areas. The new method takes the entire object as a hash instead of only the necessary tags. I'm probably a minority, but I use the `alt` tag for my banner's alt text on each post. 

post frontmatter
```
image:
  path: /assets/img/2017/02/banner.png
  alt: Travis-CI build log for my Jekyll site
  height: 400
  width: 700
```
which resulted in:
```
"image":{
  "alt":"Travis-CI build log for my Jekyll site",
  "height":400,
  "width":700,
  "url":"https://www.kyleniewiada.org/assets/img/2017/02/banner.png",
  "@type":"imageObject"
}
```

As for the author being `null`:
I do not have a test for this. But I did find a reason. Due to my author name being listed as a default value under my `_config.yml` file, Ruby never resolves my author name; only when it is in the file frontmatter. The config below no longer works with the author name.

_config.yml
```
defaults:
  - scope:
      path: _posts
    values:
      layout: post
      author: Kyle Niewiada
      seo:
        type: BlogPosting
      regenerate: true
```

### DIFF
--- a/lib/jekyll-seo-tag/json_ld.rb
+++ b/lib/jekyll-seo-tag/json_ld.rb
@@ -51,8 +51,6 @@ module Jekyll
         hash = image.dup
         hash["url"]   = hash.delete("path")
         hash["@type"] = "imageObject"
-        keepers = ["@type", "height", "width", "url"]
-        hash.keep_if { |k, _| keepers.include?(k) }
         hash
       end
 
--- a/lib/jekyll-seo-tag/json_ld.rb
+++ b/lib/jekyll-seo-tag/json_ld.rb
@@ -51,6 +51,8 @@ module Jekyll
         hash = image.dup
         hash["url"]   = hash.delete("path")
         hash["@type"] = "imageObject"
+        keepers = ["@type", "height", "width", "url"]
+        hash.keep_if { |k, _| keepers.include?(k) }
         hash
       end
 
--- a/lib/jekyll-seo-tag/json_ld.rb
+++ b/lib/jekyll-seo-tag/json_ld.rb
@@ -40,7 +40,7 @@ module Jekyll
         return unless author
         {
           "@type" => "Person",
-          "name"  => author["name"],
+          "name"  => author["name"] || page["author"],
         }
       end
 
--- a/lib/jekyll-seo-tag/json_ld.rb
+++ b/lib/jekyll-seo-tag/json_ld.rb
@@ -65,7 +65,7 @@ module Jekyll
             "url"   => logo,
           },
         }
-        output["name"] = author["name"] if author
+        output["name"] = author["name"] || page["author"] if author
         output
       end
 
--- a/History.markdown
+++ b/History.markdown
@@ -1,5 +1,8 @@
 ## HEAD
 
+* Convert template logic to a Liquid Drop (significant performance improvement) (#184)
+* Fix for JSON-LD validation warning for images missing required properties (#183)
+
 ## 2.2.0
 
 ### Major Enhancements
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -98,7 +98,7 @@ module Jekyll
           date = if page["seo"] && page["seo"]["date_modified"]
                    page["seo"]["date_modified"]
                  else
-                   page["last_modified_at"] || page["date"]
+                   page["last_modified_at"].to_liquid || page["date"]
                  end
           filters.date_to_xmlschema(date) if date
         end
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -97,8 +97,10 @@ module Jekyll
         @date_modified ||= begin
           date = if page["seo"] && page["seo"]["date_modified"]
                    page["seo"]["date_modified"]
+                 elsif page["last_modified_at"]
+                   page["last_modified_at"].to_liquid
                  else
-                   page["last_modified_at"] || page["date"]
+                   page["date"]
                  end
           filters.date_to_xmlschema(date) if date
         end
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -97,10 +97,8 @@ module Jekyll
         @date_modified ||= begin
           date = if page["seo"] && page["seo"]["date_modified"]
                    page["seo"]["date_modified"]
-                 elsif page["last_modified_at"]
-                   page["last_modified_at"].to_liquid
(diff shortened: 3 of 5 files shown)


aav7fl:
How about this attempt to handle the image hash object? I wanted to take a swing at fixing this, but I'm not sure if this is the approach you would like to follow.

lib/jekyll-seo-tag/json_ld.rb
```diff
@@ -51,6 +51,8 @@ def json_image
         hash = image.dup
         hash["url"]   = hash.delete("path")
         hash["@type"] = "imageObject"
+        keepers = ["@type", "height", "width", "url"]
+        hash.keep_if { |k, _| keepers.include?(k) }
```
aav7fl: Here is my attempt to handle the extra tags that exist in the image hash and not using them for the output.

benbalter:
This is a tricky question about what we want the plugin to be/do:

* Option 1: Take the user's input, turn it into JSON, and trust they know better than we do. Here the onus is on the user to provide valid JSON-LD metadata

* Option 2: Take the user's input, wrangle it into JSON-LD to the best of our ability, and believe we know better than they do. Here the onus is on us to provide valid JSON-LD metadata.

My gut says option 1 would be the smarter route in the long run (you give us valid input, we give you valid output), versus a garbage in, gold out approach, but am up to be convinced otherwise.

aav7fl:
I do feel option 1 in this case would be better. We can't possibly know what every user *might* do in the future, or their setup now-- even if it doesn't fit my box.

Do you know if any other GH Pages gems use the image block at all? 

How should I handle the author issue with the Jekyll supported default values method? (Using default values for `author` in the `_config.yml`.

benbalter:
Alternatively, if the spec is stable and well defined (recent discussions suggest otherwise) we can whitelist specific keys.

aav7fl:
Personally, I think we should wait a little while before whitelisting any. The spec needs a bit of an update first. 

As for the `name` being `null` (under the `publisher` object, and `author` object), it definitely looks like the front-matter default for `author` is being ignored when placed in the `_config.yml`.  But not `title`/`headline`.

lib/jekyll-seo-tag/json_ld.rb
```diff
@@ -40,7 +40,7 @@ def json_author
         return unless author
         {
           "@type" => "Person",
-          "name"  => author["name"],
+          "name"  => author["name"] || page["author"],
```
aav7fl: Is this an *alright* kind of way to handle this? It works by looking at the page and fixes my null case issues. (Just like how the title is grabbed in the drop file).

lib/jekyll-seo-tag/json_ld.rb
```diff
@@ -63,7 +65,7 @@ def publisher
             "url"   => logo,
           },
         }
-        output["name"] = author["name"] if author
+        output["name"] = author["name"] || page["author"] if author
```
aav7fl: Other special handling here.

aav7fl:
The `null` author name case is now being handled. Removed the special handling of the extra image tags. I feel it is out of scope for now.

As for the `null` name handling; I'm not sure if this is how you would like it. I'm following the method used to get `headline` from the `drop.rb` file. But this is in the `json_ld.rb` generator. Not sure if I'm out of scope by calling `page["author"]` here. 

Is this approach alright?

lib/jekyll-seo-tag/json_ld.rb
benbalter: Theoretically, the `author` object should resolve to the actual author (be it stored as hash or a string, in the post, in data, or in config), and consistently return a hash, so we don't have to do this.

benbalter:
> As for the name being null (under the publisher object, and author object), it definitely looks like the front-matter default for author is being ignored when placed in the _config.yml. But not title/headline.

Theoretically, both title/headline and author should use the same [`post` object](https://github.com/jekyll/jekyll-seo-tag/blob/master/lib/jekyll-seo-tag/drop.rb#L188), which calls `to_liquid` and thus should have defaults. 

I'm not saying it's not a bug, I'm saying it *shouldn't* be. For the problem you're describing here, can you provide steps that I can use to reproduce it locally?

aav7fl:
(This can be recreated with the latest plugin from the master branch: f048342e2296f6a8a915b27eef3326d28e02d7a0) 

Create a post with the following front matter:

`2017-02-22-example.md`
```
---
title: "TitleTest"
date: "2017-02-22 5:34"
published: true
description: "LongDescription"
---
```

Now in the `_config.yml`:
```
#Defaults
defaults:
  - scope:
      path: _posts
    values:
      layout: post
      author: Kyle Niewiada
      seo:
        type: BlogPosting
      regenerate: true
```

The JSON-LD output will list the value for `name` as `null` in the `author` and `publisher` objects. Before the large drop merge, this worked fine. Now the name only works when using the change in my PR.

This *should* make it locally reproducible.

benbalter:
@aav7fl added an explicit test for respecting front matter via 812eb22.

I recreated the above example in a new site (e.g., with nothing else, no other plugins, etc.) and got the following:

```html
<script type="application/ld+json">
{"@context":"http://schema.org","@type":"BlogPosting","headline":"TitleTest","author":{"@type":"Person","name":"Kyle Niewiada"},"datePublished":"2017-02-22T05:34:00-05:00","dateModified":"2017-02-22T05:34:00-05:00","description":"LongDescription","mainEntityOfPage":{"@type":"WebPage","@id":"/2017/02/22/example.html"},"url":"/2017/02/22/example.html"}</script>
```

Which appears to be the expected output?

aav7fl:
Odd... it looks like I need to debug further to see what is interfering. 

`https://github.com/aav7fl/aav7fl.github.io/tree/SEOTest` It's definitely causing a problem here (ignoring the dateModified build issue).

aav7fl:
Got it. More needs to be added. There is a conflict with the way I am using the twitter/authors.

Add to `_data/authors.yml`
```
Kyle Niewiada:
  image: /assets/img/kyle.jpg
  twitter: CrypticCitrus
```
 It finds the author here, grabs it, can't read it later, and evaluates to `null`.

benbalter:
> There is a conflict with the way I am using the twitter/authors.

Ahh thanks. Found it. Check out https://github.com/jekyll/jekyll-seo-tag/pull/192.

aav7fl:
Thanks! I'll close this PR now.

PR #192, opened 2017-04-13
Fix for including name when author is passed by reference

benbalter:
If author metadata is stored in a datafile without an explicit name, the previous behavior was to fall back to the author name passed as the data key. This PR preserves that behavior, and adds tests to ensure it continues to work going forward.

Typically, you'd add `author: benbalter` to a post's metadata, and then in in `_authors.yml` have a `benbalter` key, with values like `name: Ben Balter`.

It was possible, however to have `author: Ben Balter` in your post, and then have a `Ben Balter` key in your author's file (without duplicating the name).

This PR makes it such that if a `name` key does not exist in the resolved author hash, it uses the lookup key (e.g., `benbalter` in the first example and `Ben Balter` in the second) as the name.

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -226,6 +226,7 @@ module Jekyll
       def author_hash(author_string)
         if site.data["authors"] && site.data["authors"][author_string]
           hash = site.data["authors"][author_string]
+          hash["name"]    ||= author_string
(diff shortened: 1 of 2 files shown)


aav7fl:
This fixes #190 for me. Thanks!

aav7fl:
@benbalter Do you think this can bring this into master and the gem updated?

Issue #196, opened 2017-04-26
Builds failing with "Liquid Exception: no implicit conversion of String into Integer"


axilleas:
After updating the `jekyll-seo-tag` to `2.2.1`, the following error appears:

```
$ bundle install --path vendor
Fetching gem metadata from https://rubygems.org/...........
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies...
Installing public_suffix 2.0.5
Installing colorator 1.1.0
Installing ffi 1.9.18 with native extensions
Installing forwardable-extended 2.6.0
Installing sass 3.4.23
Installing rb-fsevent 0.9.8
Installing kramdown 1.13.2
Installing liquid 3.0.6
Installing mercenary 0.3.6
Installing rouge 1.11.1
Installing safe_yaml 1.0.4
Using bundler 1.14.6
Installing addressable 2.5.1
Installing rb-inotify 0.9.8
Installing pathutil 0.14.0
Installing jekyll-sass-converter 1.5.0
Installing listen 3.0.8
Installing jekyll-watch 1.5.0
Installing jekyll 3.4.3
Installing jekyll-seo-tag 2.2.1
Bundle complete! 2 Gemfile dependencies, 20 gems now installed.
Bundled gems are installed into ./vendor.
$ bundle exec jekyll build -d public
Configuration file: /builds/open-leadership/curriculum/_config.yml
            Source: /builds/open-leadership/curriculum
       Destination: public
 Incremental build: disabled. Enable with --incremental
      Generating... 
  Liquid Exception: no implicit conversion of String into Integer in /_layouts/default.html
jekyll 3.4.3 | Error:  no implicit conversion of String into Integer
```

Downgrading to `2.2.0` fixes this issue. Reported in https://gitlab.com/pages/jekyll/issues/15. Test repo in  https://gitlab.com/open-leadership/curriculum.

Thanks!

benbalter:
Fix in https://github.com/jekyll/jekyll-seo-tag/pull/197.

RohitK89:
@benbalter This is happening for me with multiple authors on a single post.
Disabling the plugin gets the build to succeed.

I'm specifying multiple authors as follows:
```
---
...
authors:
  - foo
  - bar
---
```

PR #197, opened 2017-04-26
Guard against arrays in subhashes

benbalter:
If `site.social` or `page.seo` are arrays, strings, or anything other than a Hash, things will 💥 . This PR creates a helper method to safely traverse subhashes, even when they are not actually hashes.

Fixes https://github.com/jekyll/jekyll-seo-tag/issues/196.


axilleas:
Thank you! ❤️

Issue #198, opened 2017-04-26
Liquid Exception: undefined method `absolute?' for nil:NilClass


mmistakes:
Came across this error testing the latest versions `2.2.1` on a Windows 7 box (`2.2.0` works fine). Looks like the problem is in [drop.rb](https://github.com/jekyll/jekyll-seo-tag/blob/master/lib/jekyll-seo-tag/drop.rb#L208)

1. Placed `{{ seo }}` in `_layouts/default.html`
2. `bundle exec jekyll build`

```
Liquid Exception: undefined method `absolute?' for nil:NilClass in /_layouts/d
efault.html
C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/jekyll-seo-tag-2.2.1/lib/jekyll-seo-tag/d
rop.rb:208:in `absolute_url?': undefined method `absolute?' for nil:NilClass (No
MethodError)
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/jekyll-seo-tag-2.2.1/lib/jek
yll-seo-tag/drop.rb:166:in `image'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/jekyll-3.4.3/lib/jekyll/drop
s/drop.rb:52:in `public_send'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/jekyll-3.4.3/lib/jekyll/drop
s/drop.rb:52:in `[]'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/liquid-3.0.6/lib/liquid/cont
ext.rb:201:in `lookup_and_evaluate'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/liquid-3.0.6/lib/liquid/vari
able_lookup.rb:48:in `block in evaluate'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/liquid-3.0.6/lib/liquid/vari
able_lookup.rb:38:in `each_index'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/liquid-3.0.6/lib/liquid/vari
able_lookup.rb:38:in `evaluate'
        from C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/liquid-3.0.6/lib/liquid/cont
ext.rb:168:in `evaluate'
[...]

PR #199, opened 2017-04-26
Guard against invalid or missing URLs

benbalter:
Fixes https://github.com/jekyll/jekyll-seo-tag/issues/198.


benbalter:
@DirtyF what would you expect the behavior to be? My intention here was to prevent the build from failing due to invalid or missing URLs. If a user provides an invalid URL right now, we output it to the best of our ability, even if it's wrong (or isn't live, etc.).

DirtyF:
@benbalter OK then it's fine, you're right people are responsible for what they enter and can rely on tools like html-proofer and CI to prevent invalid paths.

PR #204, opened 2017-05-02
Handle page title consisting of only whitespace

acoffman:
Hi, thanks for maintaining this plugin!

This is a bit of an edge case, but currently, this breaks if `page['title']` is set but consists of only whitespace.

Basically ` @page_title ||= format_string(page["title"] || site_title)` ends up setting `@page_title` to `nil` because `page["title"]` is defined, but calling `format_string` on it returns `nil`.

Then down in the `#title` method, this condition: `if page["title"] && site_title` will evaluate to `true` but then blow up because it tries to concatenate `page_title` which is `nil`.

This should correct that!

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -33,7 +33,11 @@ module Jekyll
 
       # Page title without site title or description appended
       def page_title
-        @page_title ||= format_string(page["title"] || site_title)
+        @page_title ||= if page["title"] && !page["title"].empty?
+                          format_string(page["title"])
+                        else
+                          format_string(site_title)
+                        end
       end
 
       # Page title with site title or description appended


lib/jekyll-seo-tag/drop.rb
```diff
@@ -33,7 +33,11 @@ def site_title
 
       # Page title without site title or description appended
       def page_title
-        @page_title ||= format_string(page["title"] || site_title)
+        @page_title ||= if page["title"] && !page["title"].empty?
```
benbalter: This can be simplified to `!page["title"].to_s.empty?`

acoffman:
Its a good question! I was helping another user debug the error and they had specified a blank title; we ultimately worked around it by removing the key entirely. I can see there being a need to have a blank title in cases where its a required field of another plugin or if you want something generated in your template, but without text. I suspect in this case we'd just want to treat it as not present.

Its admittedly a pretty far edge case however. If you think its worth allowing, I'll add the test onto this PR; otherwise feel free to close it!

benbalter:
> If you think its worth allowing, I'll add the test onto this PR

Yes please. Even if we don't want to support the edge case, we shouldn't break the build with unexpected user input.

benbalter:
Fixed via #206. Thanks for the inspiration and sleuthing @acoffman!

PR #206, opened 2017-05-05
Guard against empty title or description strings

benbalter:
An alternate approach to https://github.com/jekyll/jekyll-seo-tag/pull/204 and to address the comments on https://github.com/jekyll/jekyll-seo-tag/commit/4f80ea50b773d1995985e35a8a63806516c353c4#commitcomment-22010563, this PR defensively codes against `title` or `description` on the site or page level being an empty string.

The problem is if a value is `""`, `page["key"]` will return `""`, but `format_string(page["key"])` will return nil, meaning `format_string(page["key"]) + "s" if page["key"]` while intuitively right, will cause an unknown method or no implicit conversion error because `""` is truthy in Ruby.

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -67,9 +67,9 @@ module Jekyll
       end
 
       def description
-        @description ||= format_string(
-          page["description"] || page["excerpt"] || site["description"]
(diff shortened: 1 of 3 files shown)


benbalter:
@jekyllbot: merge +minor.

PR #211, opened 2017-05-24
use canonical_url specified in page if present

8parth:
assign the canonical url to canonical_url specified by user in page otherwise use default generated url.

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -183,7 +183,11 @@ module Jekyll
 
       def canonical_url
         @canonical_url ||= begin
-          filters.absolute_url(page["url"]).to_s.gsub(%r!/index\.html$!, "/")
+          if page["canonical_url"].present?
+            page["canonical_url"]
+          else
+            filters.absolute_url(page["url"]).to_s.gsub(%r!/index\.html$!, "/")
+          end
         end
       end
 
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -183,7 +183,7 @@ module Jekyll
 
       def canonical_url
         @canonical_url ||= begin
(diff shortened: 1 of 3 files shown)


lib/jekyll-seo-tag/drop.rb
```diff
@@ -183,7 +183,11 @@ def page_lang
 
       def canonical_url
         @canonical_url ||= begin
-          filters.absolute_url(page["url"]).to_s.gsub(%r!/index\.html$!, "/")
+          if page["canonical_url"].present?
```
benbalter: `.present?` only works in Rails. You'll want to use `to_s.empty?`

spec/jekyll_seo_tag/drop_spec.rb
```diff
@@ -624,4 +624,23 @@
       end
     end
   end
+
+  context "canonical url" do
+    context "when canonical url is specified for a page" do
+      let(:page) { make_page( { "title" => "page title", "canonical_url" => "https://github.com/jekyll/jekyll-seo-tag/"} ) }
+      let(:canonical_url) { 'https://github.com/jekyll/jekyll-seo-tag/' }
+      it "uses specified canonical url" do
+        puts subject.canonical_url
```
benbalter: Stray puts.

lib/jekyll-seo-tag/drop.rb
8parth: Didn't know that. Will update this with specified changes.

spec/jekyll_seo_tag/drop_spec.rb
8parth: Forgot to remove puts, is this proper way to test or I need to create posts fixtures? I am not sure about this part :sweat_smile:

benbalter:
@8parth this is great, and thanks for working through our feedback. I made a few minor formatting changes, but think this should be good to merge. Thank you!

PR #212, opened 2017-05-25
Move docs to their own folder

benbalter:
With #211, I feel like the README is getting a bit unweildy. I'd like to propose we move the docs into a dedicated `/docs` folder, and publish them out using GitHub Pages where they can be broken up (or browsed on GitHub, or locally).

Here's what it'd look like:

screen shot 2017-05-25 at 9 50 20 am

It'd be `jekyll.github.io/jekyll-seo-tag/`. Thoughts?


DirtyF:
A Table of Contents would be nice.

benbalter:
Thanks @dirtyF! 🌮

DirtyF:
@benbalter  🍉  please ☀️  😃

PR #213, opened 2017-05-30
feat(image): Add meta twitter image

CaptainJojo:
# Description

In documentation for twitter card we need to add meta `twitter:image`
See: [https://dev.twitter.com/cards/types/summary](https://dev.twitter.com/cards/types/summary)

# Validate

See new meta `twitter:image` in page when image.path is config

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -29,6 +29,7 @@
 
 {% if seo_tag.image %}
   <meta property="og:image" content="{{ seo_tag.image.path }}" />
+  <meta property="twitter:image" content="{{ seo_tag.image.path }}" />
   {% if seo_tag.image.height %}
     <meta property="og:image:height" content="{{ seo_tag.image.height }}" />
   {% endif %}


benbalter:
I believe this is not necessary. See https://dev.twitter.com/cards/getting-started#opengraph:

> When the Twitter card processor looks for tags on a page, it first checks for the Twitter-specific property, and if not present, falls back to the supported Open Graph property. This allows for both to be defined on the page independently, and minimizes the amount of duplicate markup required to describe content and experience.

CaptainJojo:
Yes, it's not necessary but by default twitter take the first image in page but I need to specify my image :(

benbalter:
> but I need to specify my image

If you specify it in the front matter, the plugin should output `og:image` which Twitter should use. Is that not the case?

benbalter:
> but I need to specify my image

If you specify it in the front matter, the plugin should output `og:image` which Twitter should use. Is that not the case?

CaptainJojo:
It's ok, I test directly in twitter og:image is take in twitter :)

PR #222, opened 2017-07-26
Configure default Twitter summary card type

### LINKED ISSUE #84: Should be able to add image to Twitter card without it being a large image
The standard [summary card](https://dev.twitter.com/cards/types/summary) has a `twitter:image` property as well.

It's nice to use it for the site logo on cards that don't have large images.

BlythMeister:
Add ability at site level to not use the large twitter summary

Current rather yucky workaround is doing this:

```
{% capture seo %}
{% seo %}
{% endcapture %}

{{ seo | replace: "summary_large_image", "summary" }}
```
fix #84

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -51,7 +51,11 @@
 
 {% if site.twitter %}
   {% if seo_tag.image %}
-    <meta name="twitter:card" content="summary_large_image" />
+      {% if site.twitter.smallSummary %}
+        <meta name="twitter:card" content="summary" />
+      {% else %}
+        <meta name="twitter:card" content="summary_large_image" />
+      {% endif %}
   {% else %}
     <meta name="twitter:card" content="summary" />
   {% endif %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -60,12 +60,8 @@
 {% endif %}
 
 {% if site.twitter %}
-  {% if seo_tag.image %}
-      {% if site.twitter.smallSummary %}
-        <meta name="twitter:card" content="summary" />
-      {% else %}
-        <meta name="twitter:card" content="summary_large_image" />
-      {% endif %}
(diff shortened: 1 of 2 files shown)


Carlows:
Would be great to have this!

mattr-:
Any other objections @pathawks @DirtyF or @benbalter ?

lib/template.html
```diff
@@ -51,7 +51,11 @@
 
 {% if site.twitter %}
   {% if seo_tag.image %}
-    <meta name="twitter:card" content="summary_large_image" />
+      {% if site.twitter.smallSummary %}
```
DirtyF: What about just naming this `twitter.summary`?

benbalter: Alternatively, perhaps something like `<meta name="twitter:card" content="{{ site.twitter.card | default: "summary_large_image" }}" />` may be more intuitive to end users?

BlythMeister: I like that idea!
Is there a way to do a double default.
E.g.
Page.twitter.card
Then
Site.twitter.card
Then
Summary large image

benbalter: Yep! You can chain defaults, e.g., `| default: page.twitter.card | default: site.twitter.card...`

BlythMeister: Sweet, leave this with me...I like that 👍

BlythMeister:
Updates made, also given option for a site level image.
On my site i had to add the same image to every post/page
Post/Page takes priority as does the card size

BlythMeister:
Disclaimer: I'm not sure how to test these changes locally, so it is in untested.

I did also update the docs as part of this pr

BlythMeister:
I have just realised this is now over complex and could be done using defaults.
Please disregard these changes I will raise a clean fresh or and link to this one.

PR #225, opened 2017-08-08
Configure default Twitter summary card type (V2)

### LINKED ISSUE #84: Should be able to add image to Twitter card without it being a large image

BlythMeister:
Update to #222 with cleaner history.

Original details:

Add ability at site level to not use the large twitter summary

Current rather yucky workaround is doing this:

```
{% capture seo %}
{% seo %}
{% endcapture %}

{{ seo | replace: "summary_large_image", "summary" }}
```
fix #84

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -6,11 +6,15 @@ The SEO tag will respect any of the following if included in your site's `_confi
 * `description` - A short description (e.g., A blog dedicated to reviewing cat gifs)
 * `url` - The full URL to your site. Note: `site.github.url` will be used by default.
 * `author` - global author information (see below)
-* `twitter:username` - The site's Twitter handle. You'll want to describe it like so:
+
+* `twitter` - The following properties are available:
+  * `twitter:username` - The site's Twitter handle
+  * `twitter:card` - The site's default card type
 
   ```yml
   twitter:
     username: benbalter
+    card: summary_large_image
   ```
 
 * `facebook` - The following properties are available:
@@ -66,6 +70,5 @@ The SEO tag will respect the following YAML front matter if included in a post, 
 * `image` - URL to an image associated with the post, page, or document (e.g., `/assets/page-pic.jpg`)
 * `author` - Page-, post-, or document-specific author information (see below)
 * `lang` - Page-, post-, or document-specific language information
-* `twitter_image_small` - Page-, post-, or document-specific flag to use small twitter image
   
 *Note:* Front matter defaults can be used for any of the above values as described in advanced usage with an image example.
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -14,8 +14,8 @@ The SEO tag will respect any of the following if included in your site's `_confi
   ```yml
   twitter:
     username: benbalter
-    card: summary_large_image
-  ```
+    card: summary
+  ```
 
 * `facebook` - The following properties are available:
   * `facebook:app_id` - a Facebook app ID for Facebook insights
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -65,3 +65,6 @@ The SEO tag will respect the following YAML front matter if included in a post, 
 * `image` - URL to an image associated with the post, page, or document (e.g., `/assets/page-pic.jpg`)
 * `author` - Page-, post-, or document-specific author information (see below)
 * `lang` - Page-, post-, or document-specific language information
+* `twitterCardType` - The type of twitter card to use
+  
+*Note:* Front matter defaults can be used for any of the above values as described in advanced usage with an image example.
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -65,6 +65,6 @@ The SEO tag will respect the following YAML front matter if included in a post, 
 * `image` - URL to an image associated with the post, page, or document (e.g., `/assets/page-pic.jpg`)
 * `author` - Page-, post-, or document-specific author information (see below)
 * `lang` - Page-, post-, or document-specific language information
-* `twitterCardType` - The type of twitter card to use
+* `twitter_image_small` - Page-, post-, or document-specific flag to use small twitter image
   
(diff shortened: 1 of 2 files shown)


BlythMeister:
Sorry for all the pr open/close.
Had a right nightmare

BlythMeister:
Have seen this done with a `twitter_image_small` flag set to true or false.
What's you opinion on that instead?

BlythMeister:
example usage:
* workaround in practice: https://github.com/BlythMeister/blyth.me.uk/blob/master/_includes/head.html#L26
* default config: https://github.com/BlythMeister/blyth.me.uk/blob/master/_config.yml#L39

BlythMeister:
I have no idea why this build is failing....can someone help?

pathawks:
Failing test looks unrelated.

BlythMeister:
Weird for it to suddenly start failing though, master looks to be passing.

pathawks:
[This](https://github.com/jekyll/jekyll-seo-tag/blob/8dc1719b8185e56da3a947f5646bc4a7eb950f70/spec/jekyll_seo_tag/drop_spec.rb#L566) is the test that is failing. I don't know how it could be related to the changes in this PR.

/cc: @benbalter

benbalter:
The failing test looks like a change in either Addressable or Jekyll's invalid URL handling behavior and can likely be removed, either here, or via a separate PR and then merge in master.

BlythMeister:
Changing liquid and documentation I can do.
Ruby...not so much 😋

pathawks:
> Changing liquid and documentation I can do.
> Ruby...not so much

For what it's worth, that was exactly where I was when I started hacking on Jekyll.

That said, if I have time tonight I will look into modifying/dropping the failing test :+1:

benbalter:
> The failing test looks like a change in either Addressable or Jekyll's invalid URL handling behavior and can likely be removed, either here, or via a separate PR and then merge in master.

Fixed via https://github.com/jekyll/jekyll-seo-tag/pull/228

lib/template.html
```diff
@@ -51,7 +51,11 @@
 
 {% if site.twitter %}
   {% if seo_tag.image %}
-    <meta name="twitter:card" content="summary_large_image" />
+    {% if page.twitter_image_small or post.twitter_image_small %}
+      <meta name="twitter:card" content="summary" />
+    {% else %}
+      <meta name="twitter:card" content="summary_large_image" />
+    {% endif %}
```
benbalter: Per the discussion over in https://github.com/jekyll/jekyll-seo-tag/pull/222#discussion_r131346046, would it make more sense to implement this as something like:

```liquid
<meta name="twitter:card" content="{{ page.twitter.card | default: site.twitter.card | default: "summary_large_image" }}" />
```

That way if Twitter adds a "medium" card next year, we're still covered.

BlythMeister:
I did try this and whilst I agree it does add future scope, the usage in the consuming site doesn't look/feel as nice as a true/false.

If adding as discussed with defaults is the only way to get this in, then I guess I concede, but imo, code should be implemented for the current scope, not all possible future scope 🙂

lib/template.html
BlythMeister: I did try this and whilst I agree it does add future scope, the usage in the consuming site doesn't look/feel as nice as a true/false.

If adding as discussed with defaults is the only way to get this in, then I guess I concede, but imo, code should be implemented for the current scope, not all possible future scope 🙂

benbalter: Not strongly opinionated. If others feel strongly one way or the other, glad to support the consensus.

BlythMeister: My other theory is, with this, users don't need to know the names Twitter use, they can see in the docs here there is a true/false to make it small.

DirtyF: There are already other types of cards (`app` and `player` for now): https://dev.twitter.com/cards/types.

BlythMeister: App and player cards require other meta data info rather than just the image, do there is quite a bit more effort required.
For a simple image based card which is what this section has, there is just small or big image.

BlythMeister: Having been 6 months on this...is the only way to get this in to force the user to put the exact card type into their front matter?

To repeat why I went with this approach, Twitter has 2 card types that would work with the rest of the meta this plugin adds.
By using a Boolean option, the user doesn't need to know the exact twitter card names (which is one plus for this whole plugin, the user doesn't need to know all the meta tags)

I would appreciate a definite change to get in, or a merge as is so this can move forward please.

pathawks: The boolean seems clunky. Ie: Why `twitter_image_small: true` instead of `twitter_image_large: false`?

Let's go with `twitter.card: summary_large_image` :+1:

pathawks:
@BlythMeister Sorry this has sat for so long. Ping me when you've made the changes and I will review.

BlythMeister:
@pathawks updates made as requested.

docs/usage.md
```diff
@@ -6,11 +6,15 @@ The SEO tag will respect any of the following if included in your site's `_confi
...
-* `twitter:username` - The site's Twitter handle. You'll want to describe it like so:
+
+* `twitter` - The following properties are available:
+  * `twitter:username` - The site's Twitter handle
+  * `twitter:card` - The site's default card type
 
   ```yml
   twitter:
     username: benbalter
+    card: summary_large_image
```
DirtyF: What are the other card types available? We could use a value different from the default's one here.

BlythMeister:
@DirtyF updated to "summary" which is the smaller card

BlythMeister:
@DirtyF @benbalter @pathawks can this be put in now?

pathawks:
Awesome. Thanks for all your work on this! :+1::beers:

@jekyllbot: :shipit: +minor

PR #227, opened 2017-08-22
Ensure site.data.authors is properly formatted before attempting to retrieve author meta

benbalter:
If `site.data.authors` is an array, or if `site.data.authors[author]` is anything other than a Hash, the build will error out. 

This PR makes `Drop#author_hash` a bit more robust by ensuring `site.data.authors` and `site.data.authors[author]` is a hash before attempting to access it or add additional elements. 

The user-facing functionality should be identical, but more robust.

This problem arises when Jekyll SEO Tag is included (by default in some cases) on GitHub Pages, and the user has a `_data/authors.yml` that was not formatted to work with Jekyll SEO tag.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -12,5 +12,7 @@
       
       {{ content }}
     </div>
+    <script src="{{ "assets/javascript/anchor-js/anchor.min.js" | relative_url }}"></script>
+    <script>anchors.add();</script>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


benbalter:
Failing test fixed via https://github.com/jekyll/jekyll-seo-tag/pull/228.

benbalter:
Thanks for the 👀! :smile:

PR #228, opened 2017-08-22
Fix for image.path causing an invalid url error

benbalter:
If `image.path` is invalid, `absolute_url?` will return `nil` which is falsy. Explicitly look for `false` before we try to make the URL absolute to avoid an upstream error.

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -168,7 +168,8 @@ module Jekyll
         image["path"] ||= image["facebook"] || image["twitter"]
         return @image = nil unless image["path"]
 
-        unless absolute_url? image["path"]
+        # absolute_url? will return nil for an invalid URL
+        if absolute_url?(image["path"]) == false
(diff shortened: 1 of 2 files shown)


benbalter:
I should have been more explicit. I'm sorry. This is a fix for the failing test on `master` described in https://github.com/jekyll/jekyll-seo-tag/pull/225#issuecomment-321258155.

> Whats does : stand for here?

Jekyll used to have issues with `:`s in paths, fixed in 3.5. When we call `filters.absolute_url`, that method doesn't guard against Addressable invalid URL error when it tries to parse the URL.

This was implemented originally via https://github.com/jekyll/jekyll-seo-tag/pull/199:

> My intention here was to prevent the build from failing due to invalid or missing URLs. If a user provides an invalid URL right now, we output it to the best of our ability, even if it's wrong (or isn't live, etc.).

DirtyF:
@benbalter crystal clear, thanks for the heads up.

benbalter:
@jekyllbot: merge +minor

PR #232, opened 2017-08-24
Define path with __dir__

DirtyF:
Inspired from https://github.com/jekyll/jekyll/pull/6087 who itself took inspiration from https://github.com/rails/rails/commit/5b8738c2df003a96f0e490c43559747618d10f5f

/cc @bogdanvlviv

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,4 +1,4 @@
-$LOAD_PATH.unshift File.expand_path("../../lib", __FILE__)
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
 require "jekyll"
 require "jekyll-seo-tag"
 require "html-proofer"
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -15,11 +15,11 @@ class Drop
 ENV["JEKYLL_LOG_LEVEL"] = "error"
 
 def dest_dir
-  File.expand_path("../tmp/dest", File.dirname(__FILE__))
+  File.expand_path("tmp/dest", __dir__)
```
bogdanvlviv: It should be `File.expand_path("../tmp/dest", __dir__)`

spec/spec_helper.rb
```diff
@@ -15,11 +15,11 @@ class Drop
...
 end
 
 def source_dir
-  File.expand_path("./fixtures", File.dirname(__FILE__))
+  File.expand_path("./fixtures", __dir__)
```
bogdanvlviv: would be nice to get rid of redundant `.`

benbalter:
Ooop. Sorry @DirtyF @bogdanvlviv got excited by learning something new and moved way too quickly. Glad to address the comments in a follow up PR. Sorry for the itchy finger. 100% my bad.

DirtyF:
@bogdanvlviv Thanks for the review.
@benbalter As tests were passing, I thought this was OK. 😞

benbalter:
Since it was my mistake, I implemented both changes via cf90f90. Sorry about that.

bogdanvlviv:
@benbalter, @DirtyF Nice work! :heart:

PR #236, opened 2017-09-05
Add meta generator

DirtyF:
Following discussion in https://github.com/jekyll/jekyll/issues/6347#issuecomment-327288571

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -3,6 +3,8 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
+<meta name="generator" content="Jekyll v{{ jekyll.version }}" />
+
 {% if seo_tag.page_title %}
   <meta property="og:title" content="{{ seo_tag.page_title }}" />
 {% endif %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -3,7 +3,7 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
-<meta name="generator" content="Jekyll v{{ jekyll.version }}" />
+<meta name="generator" content="Jekyll v{{ seo_tag.jekyll_version }}" />
 
 {% if seo_tag.page_title %}
   <meta property="og:title" content="{{ seo_tag.page_title }}" />
--- a/spec/jekyll_seo_tag_integration_spec.rb
+++ b/spec/jekyll_seo_tag_integration_spec.rb
@@ -23,6 +23,11 @@ RSpec.describe Jekyll::SeoTag do
     expect(output).to match(%r!Jekyll SEO tag v#{version}!i)
   end
 
+  it "outputs jekyll version" do
+    jekyll_version = Jekyll::VERSION
+    expect(output).to match(%r!Jekyll v#{jekyll_version}!i)
+  end
+
   it "outputs valid HTML" do
     site.process
     options = {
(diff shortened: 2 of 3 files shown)


lib/template.html
```diff
@@ -3,6 +3,8 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
+<meta name="generator" content="Jekyll v{{jekyll.version}}" />
```
benbalter: Should have whitespace around the `{{`s to remain consistent.

benbalter: Based on the CI output, it may not actually be outputting the version.

pathawks:
```html
<meta name="generator" content="{{ jekyll.version | prepend:'Jekyll ' }}" />
```

?

DirtyF: You're right it doesn't.  So should it be included in the drop instead?

parkr:
Definitely ensure it's tested :+1:

lib/template.html
pathawks: Good catch! `jekyll` needs to be added to the payload.

https://github.com/jekyll/jekyll-seo-tag/blob/1b66cb1522d695c59b8753e6cdeaa6e65c16f9ca/lib/jekyll-seo-tag.rb#L46-L53

DirtyF: @pathawks Same result. It looks like in the template it does not compute `jekyll.version`

benbalter: It looks like https://github.com/jekyll/jekyll/blob/master/lib/jekyll/drops/jekyll_drop.rb will need to be added to the payload @pathawks linked to. Out of scope for this PR, but it may make more sense to start with the `site_payload` (unified payload) drop and merge in the page and our drop rather than adding assumed globals one-off.

DirtyF: As I have no idea how to deal with the `site_payload` and the `jekyll_drop` for now, I followed the current pattern used for other values and added a test and to my utter surprise it works. 💧

spec/jekyll_seo_tag_integration_spec.rb
```diff
@@ -23,6 +23,11 @@
     expect(output).to match(%r!Jekyll SEO tag v#{version}!i)
   end
 
+  it "outputs jekyll version" do
+    jekyll_version = Jekyll::VERSION
+    expect(output).to match(%r!Jekyll v#{jekyll_version}!i)
```
parkr: You can use `#{Jekyll::VERSION}` here. You don't use `jekyll_version` anywhere else so there's no need for the variable.

spec/jekyll_seo_tag_integration_spec.rb
```diff
@@ -296,6 +301,7 @@
         expected = <<-EOS
 <!-- Begin Jekyll SEO tag v#{version} -->
 <title>Foo</title>
+<meta name="generator" content="Jekyll v3.5.2" />
```
parkr: Definitely make this programmatic or else this will fail when built with any other Jekyll version. I don't believe we lock this so CI will upgrade and this will fail. Use `#{Jekyll::VERSION}` like above.

benbalter:
@jekyllbot: merge +minor.

benbalter:
Nice @DirtyF! Given my recent Drop work, will see if I can use the unified drop in a subsequent PR as described above.

PR #237, opened 2017-09-06
Pass the unified drop to the template rather than an ad-hoc payload

benbalter:
As mentioned in https://github.com/jekyll/jekyll-seo-tag/pull/236#discussion_r137122176, when we wanted access to the `jekyll` universal namespace in the template, we had to add it as a one-off. 

Users expect certain globals to exist at render time. Things like `jekyll`, `site`, and `page`. We could expose each to the template manually, or, we could use the existing Jekyll::UnifiedDrop (what Jekyll uses to render pages), and merge in our custom metadata, just as Jekyll would before rendering a template.

Practically, this means that the template now has access to the JekyllDrop in the top-level `jekyll` namespace, but it also means that if Jekyll adds additional properties to the UnifiedDrop, we'll get them for free.

My thinking is that we may want to allow users to override the template with say, `_layouts/seo-tag.html`, and providing them with a predictable, intuitive, and consistent render context could only help get us one step closer.

Finally, I realized the Jekyll::SeoTag itself had integration tests, but no unit tests, so I added a few quick unit tests to ensure the payload contained the payload we'd expect (in addition to the integration test for the Jekyll version).

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -44,12 +44,12 @@ module Jekyll
     end
 
     def payload
-      {
+      # site_payload is an instance of UnifiedPayloadDrop. See https://git.io/v5ajm
+      @payload ||= context.registers[:site].site_payload.merge({
         "page"      => context.registers[:page],
-        "site"      => context.registers[:site].site_payload["site"],
         "paginator" => context["paginator"],
         "seo_tag"   => drop,
-      }
+      })
     end
 
     def drop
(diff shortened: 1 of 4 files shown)


lib/jekyll-seo-tag.rb
```diff
@@ -44,12 +44,12 @@ def options
     end
 
     def payload
-      {
+      # site_payload is an instance of UnifiedPayloadDrop. See https://git.io/v5ajm
+      Jekyll::Utils.deep_merge_hashes(context.registers[:site].site_payload, {
```
DirtyF: Why didn't I think of that?  :flushed:

PR #246, opened 2017-11-08
Fix: Consistently use self-closing tags

localheinz:
This PR

* [x] consistently uses self-closing tags

💁‍♂️ Sometimes they are closed, sometimes they aren't - better to be consistent, I think.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -45,10 +45,10 @@
 {% endif %}
 
 {% if paginator.previous_page %}
-  <link rel="prev" href="{{ paginator.previous_page_path | absolute_url }}">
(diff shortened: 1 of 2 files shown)


DirtyF:
@localheinz This breaks the tests. I'd better close all tags even if it's optional in HTML5.

localheinz:
@DirtyF 

Ha, hang on - let me close the tags consistently and then see about making the tests pass!

DirtyF:
Thanks @localheinz for this first contribution 👍 

@jekyllbot: merge +minor

localheinz:
Thank you, @benbalter, @DirtyF, and @pathawks!

PR #250, opened 2017-11-17
Include page number in title

mpchadwick:
Per [Moz's article "Pagination: Best Practices for SEO & User Experience"](https://moz.com/blog/pagination-best-practices-for-seo-user-experience) it's not ideal to use the same title with pagination. This change follows Moz's suggestion and appends "Page X of Y for" to the beginning of the `<title>` when on page 2 of deeper.

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -182,7 +182,7 @@ module Jekyll
       end
 
       def page_number
-        return if !@context["paginator"] || !@context["paginator"]["page"]
+        return unless @context["paginator"] && @context["paginator"]["page"]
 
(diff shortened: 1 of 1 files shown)


DirtyF:
@jekyllbot: merge +minor

DirtyF:
Thanks @mpchadwick 😃

lib/jekyll-seo-tag/drop.rb
```diff
@@ -53,6 +53,12 @@ def title
             page_title || site_title
           end
         end
+
+        if page_number
+          return page_number + @title
+        end
```
pathawks: Is it intentional that we are not memoizing this?

chesterhow:
Is there a way to disable this feature?

PR #259, opened 2017-12-29
Support Jekyll Assets for images

### LINKED ISSUE #94: Integration with jekyll-assets?
Do you have any plans to provide integration with [jekyll-assets](https://github.com/jekyll/jekyll-assets) so that if somebody's using that gem with jekyll-seo-tag, they can specify the name of an image in "site.logo" or "page.image" and have everything "just work"?

saleemrashid:
Adds `image.asset` to provide a Jekyll Assets filename instead of a path to the file (if `image.path` is not provided).

Fixes #94

### DIFF
--- a/lib/jekyll-seo-tag/image_drop.rb
+++ b/lib/jekyll-seo-tag/image_drop.rb
@@ -36,15 +36,42 @@ module Jekyll
       attr_accessor :page
       attr_accessor :context
 
+      def sprockets
+        @sprockets ||= if context.registers[:site].respond_to? :sprockets
+                         context.registers[:site].sprockets
+                       end
+      end
+
+      def digest_path
+        if sprockets.nil?
+          raise "Jekyll Assets is required"
+        else
+          @digest_path ||= begin
+            asset = sprockets.find_asset! @image_hash["asset"]
+            sprockets.manifest.compile(asset.logical_path)
+
+            sprockets.prefix_url(asset.digest_path)
+          end
+        end
+      end
+
       # The normalized image hash with a `path` key (which may be nil)
       def image_hash
-        @image_hash ||= if page["image"].is_a?(Hash)
-                          { "path" => nil }.merge(page["image"])
-                        elsif page["image"].is_a?(String)
-                          { "path" => page["image"] }
-                        else
-                          { "path" => nil }
-                        end
+        return @image_hash if defined? @image_hash
+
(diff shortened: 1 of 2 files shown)


lib/jekyll-seo-tag/image_drop.rb
```diff
@@ -36,15 +36,42 @@ def path
...
 
+      def sprockets
+        @sprockets ||= if context.registers[:site].respond_to? :sprockets
+                         context.registers[:site].sprockets
+                       end
+      end
+
+      def digest_path
+        if sprockets.nil?
+          raise "Jekyll Assets is required"
```
saleemrashid: I haven't contributed to Jekyll for a long time so I'm not sure if this is the correct approach. Should I use `Jekyll.logger.warn` instead?

saleemrashid:
CI failure is due to Rubocop 0.52.0 update.

/cc @envygeeks to review Jekyll Assets code

saleemrashid:
As an alternative, I could implement this in Jekyll Assets (which would work with themes that use `post.image.path`)

benbalter:
> As an alternative, I could implement this in Jekyll Assets (which would work with themes that use post.image.path)

@saleemrashid if it's be possible to set `post.image.path` in Jekyll assets, I think that would provide a more consistent user experience for themes and end users than special casing one plugin in another.

nicholasjstock:
@saleemrashid  did you ever end up implementing this on jekyll-assets side? I am trying to solve the same problem...

saleemrashid:
@nicholasjstock This was more than two years ago, so I don't recall what happened, but I had written a plugin that I use to implement this:

```ruby
# frozen_string_literal: true

module Jekyll
  class FrontmatterImage
    def self.setup
      Jekyll::Hooks.register :documents, :pre_render do |document|
        run_with(document)
      end
    end

    def self.run_with(document)
      site = document.site
      hash = document["image"]

      update_frontmatter(site.sprockets, hash)
    end

    def self.update_frontmatter(sprockets, hash)
      return unless hash.is_a? Hash
      return if hash["asset"].nil?

      hash["path"] ||= digest_path(sprockets, hash["asset"])
    end

    def self.digest_path(sprockets, name)
      asset = sprockets.find_asset!(name)
      sprockets.manifest.compile(asset.logical_path)

      sprockets.prefix_url(asset.digest_path)
    end
  end
end

Jekyll::FrontmatterImage.setup
```

PR #276, opened 2018-03-01
Docs: Prevent GitHub Pages from processing Liquid raw tag

### LINKED ISSUE #275: Installation Documentation is Misleading.
In the installation documentation, it currently reads as though You are supposed to put the following before `</head>`:

    {% raw %}
    {% seo %}
    {% endraw %}

Since the correct way is simply including:

    {% seo %}

Shouldn't it be listed as such?

DirtyF:
fix #275


DirtyF:
@jekyllbot: :shipit:

PR #285, opened 2018-04-30
WIP: Drop support for Ruby 2.2

pathawks:
CI is going to fail, because this Gem tests with GitHub Pages, which has not yet been updated to Jekyll 3.8 which includes necessary changes to the Rubocop config.


DirtyF:
@pathawks you might want to stop testing CI with Ruby 2.1 and 2.2

PR #292, opened 2018-05-18
Target Ruby 2.3 and Rubocop 0.56.0

DirtyF:
CI may fail as GitHub Pages uses Jekyll 3.7.3 rubocop's config for now, but this has been successfully tested with Jekyll 3.8.1.


jekyll-seo-tag.gemspec
```diff
@@ -30,5 +30,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.15"
   spec.add_development_dependency "html-proofer", "~> 3.7"
   spec.add_development_dependency "rspec", "~> 3.5"
-  spec.add_development_dependency "rubocop", "0.51"
+  spec.add_development_dependency "rubocop", "0.56.0"
```
pathawks: Do we want this to be `"~> 0.56.0"`?

DirtyF: yes, because Rubocop has a tendency to add new cops that make our tests fail :)

pathawks: In patch releases?

Right now, it looks like you've locked it to `"0.56.0"`. Should we change it to `"~> 0.56.0"` to allow (ie) `"0.56.1"`?

.travis.yml
```diff
@@ -1,6 +1,4 @@
 rvm:
-  - 2.1
-  - 2.2
```
pathawks: Should we also add `spec.required_ruby_version = ">= 2.3.0"` to the `gemspec`?

pathawks:
@jekyllbot: :shipit: +dev

PR #293, opened 2018-05-18
Release :gem: 2.5.0

pathawks:
I am not an owner, so I can not publish a new Gem.

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -1,4 +1,4 @@
-## HEAD
+## 2.5.0 / 2018-05-18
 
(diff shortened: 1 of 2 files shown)


benbalter:
@pathawks, @DirtyF Released 2.5.0, but it looks like I'm the only maintainer, which is no bueno. If you [email me](mailto:ben@balter.com) your Rubygems username, I'd be glad to add you (unless there's another/better way?).

pathawks:
We should add this gem to [**jekyll/maintainers**](https://github.com/jekyll/maintainers/blob/master/gems.yml). I am not able to open a PR in that repository, but @DirtyF can.

Issue #265, opened 2018-01-26
Duplicate meta descriptions with pagination


ouyi:
When [pagination](https://jekyllrb.com/docs/pagination/) is enabled, the description tags are identical for the paginated pages. Google Search Console lists this as an issue under Search Appearance => HTML Improvements => Pages with duplicate meta descriptions .

Is it possible to include the page number in the description when pagination is enabled?

For example:
```
<meta name="description" content="My description - Page 2 of 3">
<meta property="og:description" content="My description - Page 2 of 3">
...
```

benbalter:
If anyone is interested in taking this issue on, I believe the relevant change would need to be made [here](https://github.com/jekyll/jekyll-seo-tag/blob/master/lib/jekyll-seo-tag/drop.rb#L40-L43).

alokpant:
@benbalter Can I look into this? So to make sure that I understand:

when pagination is enabled it should show something like this:
```
<meta name="description" content="My description - Page 2 of 3">
<meta property="og:description" content="My description - Page 2 of 3">
```

if not

```
<meta name="description" content="My description">
<meta property="og:description" content="My description">
```

benbalter:
@alokpant Yep. Exactly that. https://github.com/jekyll/jekyll-seo-tag/pull/250 may be a good place to start as it looks like it solved a similar problem.

alokpant:
Cool thank you. I will look into the reference and try to solve this too.

alokpant:
@benbalter bit confused, shouldn't change be required [here](https://github.com/jekyll/jekyll-seo-tag/blob/46ad253d4de8d8064b4156a9db530399bdd9ec11/lib/jekyll-seo-tag/drop.rb#L36) rather than [here](https://github.com/jekyll/jekyll-seo-tag/blob/46ad253d4de8d8064b4156a9db530399bdd9ec11/lib/jekyll-seo-tag/drop.rb#L40-L43) as you provided a clue above. As task is to change meta-description.

How can I test it before submitting PR?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

[...]

PR #297, opened 2018-05-24
Update Page description to include page number

alokpant:
Issue: https://github.com/jekyll/jekyll-seo-tag/issues/265
Add page number to description if it is not first page.
First Page Description: 'My Description'
Other pages: 'My Description - Page 1 of 2'

### DIFF
--- a/History.markdown
+++ b/History.markdown
@@ -1,3 +1,9 @@
+## HEAD
+
+### Bug Fixes
+
+  * Update url of schema website (#296)
+
 ## 2.5.0 / 2018-05-18
 
   * Docs: Prevent GitHub Pages from processing Liquid raw tag (#276)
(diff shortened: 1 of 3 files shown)


alokpant:
@benbalter Earlier version of code works. Is there a way to test code I have written works as well? 

Should I also try to include spec for new code that I have added? I tried to write it but wasn't successful. I looked into helper method to see if I can reuse any, there are methods on creating single pages but none for multiple ones. If specs are needed, I would probably need direction or help around how to how to create multiple pages.

DirtyF:
@alokpant 

I think you should be able to pass options to the `make site` method. here you would need to pass pagination options, build the site, then check that the descriptions of the paginated pages are OK.

alokpant:
@DirtyF I tried it last week, couldn't manage to do it. I will try it this week.

alokpant:
@DirtyF I tried to pass paginator option as part of config, but @context['paginator'] was still nil.
```
        let(:config) {
          {
            :description => "My site description",
            :paginator => {
              page: 2,
              total_pages: 3
            }
          }
        }
```

Can you please suggest what am I missing?

PR #300, opened 2018-07-05
Add ahrefs site verification tag

tongrhj:


benbalter:
Thanks for this. Do you mind providing a bit more information about what problem you're trying to solve?

DirtyF:
@benbalter https://ahrefs.com/  

We should just stick with major search engines.

PR #306, opened 2018-10-03
Do not cache the drop payload for SeoTag

ashmaroli:
For compatibility with Jekyll 4.0.. and current `jekyll:master`..

### DIFF
--- a/lib/jekyll-seo-tag.rb
+++ b/lib/jekyll-seo-tag.rb
@@ -56,7 +56,20 @@ module Jekyll
     end
 
     def drop
-      @drop ||= Jekyll::SeoTag::Drop.new(@text, @context)
+      # TODO: Modify following comment once Jekyll 4.0 ships.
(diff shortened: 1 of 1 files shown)


yyyc514:
Since we have the conditional is there a reason this is hung and not getting merged into master?

ashmaroli:
> is there a reason this is hung

Perhaps the maintainers are waiting for Jekyll core to finalize whether the said "caching Liquid templates" enhancement would be included in the final release in order to avoid unnecessary code-churn...

Additionally, the conditional is a hack and as of now, the line
`if context.registers[:site].liquid_renderer.respond_to?(:cache)` would be executed for every call to the `drop` method, affecting performance in sites not willing to upgrade to  Jekyll 4.x..

Lots of decisions to take..

DirtyF:
@jekyllbot: merge +fix

PR #312, opened 2018-10-14
Optional generator and comment in template.html

maveille:
Comments and Generator tag can be optional


maveille:
i will try in local and push request after

Issue #307, opened 2018-10-09
Feature Request: Allow removing generator meta tag


philippemarlin:
Hi,
Is it possible with option to remove this line ?
 `<meta name="generator" content="Jekyll v3.3.1" />`

ashmaroli:
Why do you want to remove it??

philippemarlin:
it's a footprint in Se0

ashmaroli:
More details, please.
How is this "footprint" hampering your site's SEO?

philippemarlin:
it's a footprint will can trace a website technology for private bl0g

philippemarlin:
@DirtyF i know that is a minority (source you can use https://www.wappalyzer.com/ ). But if all websites who push to your money site with same footprint is a danger. I do wp jekyll homemade and more but if the footprint are more removed is great.

ashmaroli:
Closing as this has been marked as "wont-fix". Will reopen if a better use case can be demonstrated.

PR #313, opened 2018-10-14
Optional meta generator & comments

maveille:
Meta generator and comments can be optional with {% seo footprint=false %} param

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -12,6 +12,16 @@ If for some reason, you don't want the plugin to output `<title>` tags on each p
 ```
 <!-- {% endraw %} -->
 
+### Disabling `generator tag and comments` output
+
+If for some reason, you don't want the plugin to output `<!-- Begin Jekyll SEO tag v2.5.0 --> <meta name="generator" content="Jekyll vx.x.x" /><!-- End Jekyll SEO tag -->` tags on each page, simply invoke the plugin within your template like so:
+
+<!-- {% raw %} -->
+```
+{% seo footprint=false %}
+```
+<!-- {% endraw %} -->
+
 ### Author information
 
 Author information is used to propagate the `creator` field of Twitter summary cards. This should be an author-specific, not site-wide Twitter handle (the site-wide username be stored as `site.twitter.username`).
--- a/History.markdown
+++ b/History.markdown
@@ -7,6 +7,7 @@
 ### Development Fixes
 
   * Lint Ruby code with rubocop-jekyll gem (#302)
+  * deps: bump rubocop-jekyll to 0.3.0 (#316)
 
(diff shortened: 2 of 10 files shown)


maveille:
Anyone can help ? i dont know why its write that dont work, i have test and it is

docs/advanced-usage.md
```diff
@@ -12,6 +12,16 @@ If for some reason, you don't want the plugin to output `<title>` tags on each p
 ```
 <!-- {% endraw %} -->
 
+### Disabling `generator tag and comments` output
+
+If for some reason, you don't want the plugin to output `<!-- Begin Jekyll SEO tag v2.5.0 --> <meta name="generator" content="Jekyll vx.x.x" /><!-- End Jekyll SEO tag -->` tags on each page, simply invoke the plugin within your template like so:
```
benbalter: > If for some reason

Per @ashmaroli in https://github.com/jekyll/jekyll-seo-tag/issues/307#issuecomment-428299857, I believe the maintainers' consensus was that this feature should not be implemented within the project absent a compelling use case to justify the necessary added complexity.

DirtyF:
Closing as wontfix.

pongstr:
@DirtyF at least state a reason why it's labeled as **wont fix**

ashmaroli:
@pongstr @maveille Why is the presence of "the footprint" such an issue..?? Why are you so driven on having it not rendered..?

maveille:
Like they dont want my push, i have clone for me so if you want you can see on my repo @pongstr

pongstr:
@ashmaroli

> Why is the presence of "the footprint" such an issue..?? 
> Why are you so driven on having it not rendered..?

I'm wondering what makes you think that **I'm driven to not having it rendered**? The 👎 is for closing as `wontfix` without a reason, so please don't assume.

Personally, it doesn't bother me however when a request from the company I work for asked me to have it remove, i would like to have a valid reason to my client and also understand where the source authors are coming from.

ashmaroli:
I understand that clients can be demanding.
But the reason this was marked as `wont-fix` because *the maintainers of this project and myself don't see how having these "footprints" in an HTML page is a source of concern.*

Adding `{% if %}` blocks (and other Liquid) is a complexity that will have an impact on the overall build times (more so in large site with numerous posts, documents and pages). So, the maintainers won't accept this change unless the community provides a reason to validate the added complexity.

That said, users are free to fork the project, customize it and use the fork in their project.

pongstr:
> That said, users are free to fork the project, customize it and use the fork in their project.

No worries buddy, done that already. I just thought there's a profound reason behind. IMO, one more if doesn't really make significant `-/+` to build time. But that's okay, I get you.

Issue #298, opened 2018-05-31
Recommend Adding <html prefix=...> In The Docs


ilyaigpetrov:
Open Graph documentation suggests adding `<html prefix="og: http://ogp.me/ns#">`.  
If it is the recommended way by Open Graph then I suggest adding this prefix to your documentation.

benbalter:
That's a great suggestion @ilyaigpetrov. As you implied, it would have to be implemented by the template, so docs are the right route to go. Would you be interested in submitting a pull request with that change?

ilyaigpetrov:
@benbalter Requested a pull. Let me know if my English is not perfect or if you would prefer putting it into another document.

kennita01:
-8-

kennita01:
<8>

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

PR #314, opened 2018-10-18
Suggest adding Open Graph prefix in the docs

DirtyF:
Fix # #298

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -22,3 +22,14 @@ If you are using a Jekyll version less than `3.5.0`, use the `gems` key instead 
   {% seo %}
   ```
 <!-- {% endraw %} -->
+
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -22,3 +22,14 @@ If you are using a Jekyll version less than `3.5.0`, use the `gems` key instead
...
   ```
 <!-- {% endraw %} -->
+
+In the same template(s) add a `prefix="og: http://ogp.me/ns#"` attribute to its `<html>` tag:
+
+<!-- {% raw %} -->
+```html
+<!DOCTYPE html>
+<html lang="{{ site.lang | default: "en-US" }}" prefix="og: http://ogp.me/ns#">
+  <head>                                    <!--^ Added. -->
```
DirtyF:
```suggestion
  <head>
```

DirtyF: This comment can be deleted.

PR #319, opened 2018-12-06
Correct RuboCop offenses in spec files

ashmaroli:
`rubocop-jekyll-0.4.0` allows detecting RuboCop Offenses in spec files..


DirtyF:
@jekyllbot: merge +dev

PR #324, opened 2019-01-04
Adding possibility to change pagination message by config file

AlekseySpiridonov:
Default message can by overriten by "paginator_message" parameter in _config.yml

This change allows you to change messages for localize the site to other languages.


AlekseySpiridonov:
I added information about this change to **docs/advanced-usage.md**

jekyll-seo-tag.gemspec
```diff
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "jekyll", "~> 3.3"
-  spec.add_development_dependency "bundler", "~> 1.15"
+  spec.add_development_dependency "bundler", "~> 2.0.1"
```
DirtyF: We don't want to force bundler 2 for now, you can remove the version constraint for the tests to pass.

AlekseySpiridonov: Thanks.

AlekseySpiridonov:
I am working on tests and changes for build.  
I will update this PR.

AlekseySpiridonov:
I am working on tests. It takes some time. I will update this PR.

ashmaroli:
@AlekseySpiridonov Testing this particular change with RSpec *appears to be hard* because `@context["paginator"]` is defined by another plugin (`jekyll-paginate`) which is not loaded by default. But, if you go through the spec file and the spec_helper file, you'll realize that things are actually quite simple:
```ruby
  context "pagination" do
    let(:context) do
      make_context(
        { :page => page, :site => site },
        "paginator" => { "page" => 2, "total_pages" => 10 }
      )
    end

    it "renders the correct page number" do
      expect(subject.send(:page_number)).to eq("Page 2 of 10 for ")
    end
  end
```
The above will pass for the default message. Once that's done, you need to test for the custom message as well:
```ruby
  context "pagination" do
    let(:context) do
      make_context(
        { :page => page, :site => site },
        "paginator" => { "page" => 2, "total_pages" => 10 }
      )
    end

    it "renders the correct page number" do
      expect(subject.send(:page_number)).to eq("Page 2 of 10 for ")
    end

    # nested context
    context "with custom pagination message" do
      let(:config) { ... }

      it "" do
      end
    end
```

AlekseySpiridonov:
@ashmaroli I added tests but it always failed.

Commit - https://github.com/jekyll/jekyll-seo-tag/pull/324/commits/fbecf19e822b8d5b11b2031b5d35733835d42bc1

Logs:
```
Failures:

  1) Jekyll::SeoTag::Drop pagination renders the correct page number
     Failure/Error: paginator_message ||= @site["paginator_message"] || "Page #{current} of #{total} for "
     
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/jekyll-seo-tag/drop.rb:192:in `page_number'
     # ./spec/jekyll_seo_tag/drop_spec.rb:494:in `block (3 levels) in <top (required)>'

  2) Jekyll::SeoTag::Drop with custom pagination message renders the correct page number
     Failure/Error: let(:config) {{ "paginator_message" => "#{current} of #{total}" }}
     
     NameError:
       undefined local variable or method `current' for #<RSpec::ExampleGroups::JekyllSeoTagDrop::WithCustomPaginationMessage:0x00007f9f7f2a0d20>
     # ./spec/jekyll_seo_tag/drop_spec.rb:499:in `block (3 levels) in <top (required)>'
     # ./spec/jekyll_seo_tag/drop_spec.rb:7:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:35:in `make_page'
     # ./spec/jekyll_seo_tag/drop_spec.rb:6:in `block (2 levels) in <top (required)>'
     # ./spec/jekyll_seo_tag/drop_spec.rb:8:in `block (2 levels) in <top (required)>'
     # ./spec/jekyll_seo_tag/drop_spec.rb:10:in `block (2 levels) in <top (required)>'
     # ./spec/jekyll_seo_tag/drop_spec.rb:502:in `block (3 levels) in <top (required)>'

Finished in 1.12 seconds (files took 1.18 seconds to load)
202 examples, 2 failures
```

Could you please check my commit?  
Thanks.

ashmaroli:
@AlekseySpiridonov I knew it would fail. That's the beauty of test-driven-development.
I did not tell it to you beforehand because you would not be able to understand why it would fail..

The fact is that while your implementation works, technically it is not the proper way.
`paginator_message.sub("#\{current\}", current.to_s).sub("#\{total\}", total.to_s)` is just a hack. Additionally, there is one other mistake in your implementation.
  - `||=` is not what you think it is doing and is useless in the current context.
  - Instead of `gsub`-ing placeholders, use a proper "string template"

So, the `:page_number` method ought to be something like:
```ruby
      def page_number
        return unless @context["paginator"] && @context["paginator"]["page"]

        current = @context["paginator"]["page"]
        total = @context["paginator"]["total_pages"]
        paginator_message = site["paginator_message"] || "Page %<current>s of %<total>s for "

        format(paginator_message, :current => current, :total => total) if current > 1
      end
```
`"Page %<current>s of %<total>s for "` is the string-template which will "rendered" by the `format()` call in the succeeding line.

There's one other teeny mistake in your test as well: your custom-message context is not nested under the default-message context. Perhaps, my example wasn't clear enough.
It should be something like:
```ruby
context "..." do
  ...
  context "..." do
    ...
  end
end
```
So, the test will be something like:

```ruby
context "pagination" do
  let(:context) do
    ...
  end

  it "renders.." do
    ...
  end

  # nested context
  context "with custom message..." do
    let(:config) { { "paginator_message" => "%<current>s of %<total>s" }  }

    it "renders..." do
      ...
    end
  end
end
```
Now what you have to do is:
  - Correct the test
  - Correct your implementation
  - Correct the documentation asking users to provide the "string template" that contains two variables `current` and `total` that will be replaced as necessary.

ashmaroli:
*TODO: Merge in `master` once #325 gets merged*

ashmaroli:
@jekyll/plugin-core Should we nest the `paginator_message` key under a plugin-specific config key..? If yes, what would the key be..?

benbalter:
Would it make sense to change the message (in all cases) to "Page ___ / ____" and then just allow "page" to be localized, rather than going down the path of string interpolation? If we are going that route, would a sprintf style `%d` make more sense/be a cleaner implementation than matching pseduo-ruby in a string?

ashmaroli:
> would a sprintf style `%d` make more sense/be a cleaner implementation than matching pseduo-ruby in a string?

The current approach uses `Kernel#format`. Isn't that superior over `Kernel#sprintf`?
Please excuse my naivety, but could you explain how using `sprintf` would be a cleaner impl.. and what you meant by "matching pseudo-ruby in a string"..??

AlekseySpiridonov:
Do you need some additional changes from my side or we are waiting review?

berserk24:
I need this change. +1 for merge this pool request.

AlekseySpiridonov:
Thanks a lot!

PR #325, opened 2019-01-06
Relax version constraint on Bundler

ashmaroli:
To support Bundler v1.x series as well as its v2.x series.


DirtyF:
@ashmaroli did you see the comment from @benbalter about bundler in https://github.com/jekyll/jekyll-seo-tag/blob/e2eec18895919b2c05acc4d19354baf2234618ae/lib/jekyll-seo-tag/version.rb#L1 ?

ashmaroli:
> did you see the comment from @benbalter about bundler in

Yes. It is unrelated to this change..

DirtyF:
@jekyllbot: merge +dev

PR #330, opened 2019-02-14
Twitter Image and Title

scottwater:
I noticed Twitter is no longer pulling the image from og:image. 

This change adds the twitter:image meta tag if an image exists on the post. 

It also adds the twitter:title meta tag as well.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -54,6 +54,7 @@
 {% if site.twitter %}
   {% if seo_tag.image %}
     <meta name="twitter:card" content="{{ page.twitter.card | default: site.twitter.card | default: "summary_large_image" }}" />
+    <meta property="twitter:image" content="{{ seo_tag.image.path }}" />
   {% else %}
     <meta name="twitter:card" content="summary" />
   {% endif %}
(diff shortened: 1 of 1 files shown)


DirtyF:
Just tested https://jekyllrb.com/news/2018/11/04/jekyll-3-8-5-released/ on https://cards-dev.twitter.com/validator, and it does show the `og:image`.

scottwater:
@DirtyF Yes, I can see that image is working.  My guess is there is something going on with caching or priority in the image look up. I had an example ready to send you which showed the how my fix worked, but when I double checked Twitter did pick up the og:image but it was missing the first time. 

I will see if I can create a more definitive demo. 

Having said that, either way, it still feels like a good idea to set the twitter:image and twitter:title. 

Thanks,
Scott

scottwater:
I have not been able to duplicate this issue again, but I still believe these changes are helpful long term.

mattr-:
OpenGraph is listed as a fallback on twitter's site so I'm happy to see us provide first class support for this. Especially since they keep breaking it. 😐

@jekyllbot: merge +minor

PR #336, opened 2019-03-22
Test against Jekyll 4.x

benbalter:
Assuming tests pass, this PR will allow the plugin to be used with Jekyll 4.x.

I _believe_ based on the blame, the Pages gem was only required to make local previewing of the documentation easier (but isn't necessary for the plugin itself, tests, etc.


ashmaroli:
The documentation for RuboCop's `Lint/ToJSON` cop notes that:
> Since `JSON#generate` allows for an optional argument, your method should too.

I think that's reasonable.
But the question is, should that be done in this PR or a separate commit to `master`?

DirtyF:
@jekyllbot: merge +dev

PR #340, opened 2019-04-07
Cache formatted strings in SeoTag::Drop instances

ashmaroli:
This optimization is to avoid running through `[:markdownify, :strip_html, :normalize_whitespace, :escape_once]` multiple times for the same `string` argument (from multiple pages) and instead serve the cached output from the first run.


ashmaroli:
Profiling the change shows negative gain.
Therefore closing..

PR #341, opened 2019-04-08
Generate SeoTag::Filters instances as necessary

ashmaroli:
Out-of-the-box, `SeoTag::ImageDrop` is tightly linked to an instance of `SeoTag::Drop`. i.e., `SeoTag::ImageDrop` is only initialized via an instance of `SeoTag::Drop` (with the same `@context` used by `SeoTag::Drop`, and only when `SeoTag::Drop#image` is invoked).

So, initializing the `SeoTag::ImageDrop` object with an instance of `SeoTag::Filters` used by `SeoTag::Drop` reduces unnecessary memory allocations.


ashmaroli:
This can wait for some more time..

PR #342, opened 2019-04-08
Memoize #author_hash in SeoTag::AuthorDrop

ashmaroli:
To avoid allocating a new Hash instance for every call to `SeoTag::AuthorDrop#author_hash` in every instance of `SeoTag::AuthorDrop`


DirtyF:
@jekyllbot: merge +dev

PR #343, opened 2019-04-08
Simplify conditional in SeoTag::Drop#date_modified

ashmaroli:
This removes the unnecessary checks in the conditional and instead directly evaluates the expression to short-circuit the conditional.

(*`nil.to_liquid` returns `nil`. So a `NoMethodError` won't be raised when `page["last_modified_at"]` is`nil`.*)


DirtyF (approved):

DirtyF:
@jekyll: merge +dev

PR #344, opened 2019-04-08
Cache author name in SeoTag::JSONLDDrop

ashmaroli:
Primarily because `AuthorDrop#name` is not a memoized method nor a reader method that returns the same object for every call in a given instance.
Secondarily because this avoids repetitive method calls.

### DIFF
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -33,11 +33,11 @@ module Jekyll
       end
 
       def author
-        return unless page_drop.author["name"]
+        return unless author_name
 
         {
           "@type" => "Person",
-          "name"  => page_drop.author["name"],
+          "name"  => author_name,
         }
       end
 
@@ -61,7 +61,7 @@ module Jekyll
             "url"   => logo,
           },
         }
-        output["name"] = page_drop.author.name if page_drop.author.name
+        output["name"] = author_name if author_name
         output
       end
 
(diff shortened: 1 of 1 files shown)


parkr:
I don't see a problem with the current setup. We're deferring to the Page drop and I believe it's accessing data via the data hash on the Page so we should already be fast, right?

ashmaroli:
@parkr As of d470987 in this PR, the value of `page_drop.author["name"]` doesn't return the same object on every call. Let's walk through the flow:
1. `page_drop` is an instance of `Jekyll::SeoTag::Drop`:
https://github.com/jekyll/jekyll-seo-tag/blob/d470987d3dd617c76c5511da94f036a5233add88/lib/jekyll-seo-tag/json_ld_drop.rb#L23-L27
2. `page_drop.author` is an instance of the `AuthorDrop`:
https://github.com/jekyll/jekyll-seo-tag/blob/d470987d3dd617c76c5511da94f036a5233add88/lib/jekyll-seo-tag/drop.rb#L85-L88
3. `page_drop.author['name']` is equivalent to `page_drop.author.name` (because that's how Liquid Drops  work). `page_drop.author.name` in turn looks up `#author_hash` which is not memoized.
https://github.com/jekyll/jekyll-seo-tag/blob/d470987d3dd617c76c5511da94f036a5233add88/lib/jekyll-seo-tag/author_drop.rb#L28-L33
https://github.com/jekyll/jekyll-seo-tag/blob/d470987d3dd617c76c5511da94f036a5233add88/lib/jekyll-seo-tag/author_drop.rb#L73-L84

But, now that **#342 has been approved for merge**, this PR will eventually become a **micro-optimization** that would swap the above "walk-through-method-lookup-series" with **a single-lookup within current instance**.

PR #351, opened 2019-05-30
use example.com for example URL

coliff:
it's specially designed for examples - no need to point to http://yoursite.com :-)


DirtyF:
@jekyllbot: merge +docs

PR #356, opened 2019-08-18
Prefer site.tagline to site.description for page title

qwtel:
Using `site.description` as part of the title for pages that don't have an explicit title --- typically the landing page of a site --- leads to very long page titles. The `site.description` field is typically used for search engine or social media previews, which are 100+ characters long, making it a bad choice for the title for pages that use the description field in this way.

This PR changes the behavior of this plugin to prefer `site.tagline` over `site.description` when it is available. Sites that do not define a `site.tagline` are not affected by this. 

The name `site.tagline` was chosen as it served [the same role][1] in the Hyde theme. 

[1]: https://github.com/poole/hyde/blob/master/_includes/head.html

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -34,6 +34,10 @@ module Jekyll
         @site_title ||= format_string(site["title"] || site["name"])
       end
 
+      def site_tagline
+        @site_tagline ||= format_string site["tagline"]
+      end
+
       def site_description
         @site_description ||= format_string site["description"]
       end
@@ -43,6 +47,10 @@ module Jekyll
         @page_title ||= format_string(page["title"]) || site_title
       end
 
+      def site_tagline_or_description
+        site_tagline || site_description
+      end
+
       # Page title with site title or description appended
       # rubocop:disable Metrics/CyclomaticComplexity
       def title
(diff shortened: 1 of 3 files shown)


ashmaroli:
@qwtel Can you add a test for a scenario that doesn't have `site.tagline` defined? Thanks.

qwtel:
I didn't add one specifically b/c `site.tagline` being undefined is the case for the entire current tests suit.

qwtel:
Can this be merged now?

lib/jekyll-seo-tag/drop.rb
```diff
@@ -43,14 +47,18 @@ def page_title
...
+
       # Page title with site title or description appended
       # rubocop:disable Metrics/CyclomaticComplexity
       def title
         @title ||= begin
           if site_title && page_title != site_title
             page_title + TITLE_SEPARATOR + site_title
           elsif site_description && site_title
-            site_title + TITLE_SEPARATOR + site_description
+            site_title + TITLE_SEPARATOR + site_tagline_or_description
```
parkr: If this is the only place `site_tagline_or_description` is used, I’d suggest inlining the method. Feel free to use parentheses.

qwtel: I tried to do that, but it created a style offence: 
* `Metrics/AbcSize: Assignment Branch Condition size for title is too high`
* `Metrics/PerceivedComplexity: Perceived complexity for title is too high`

qwtel:
Will update with documentation

qwtel:
I've updated the documentation, how does it look?

qwtel:
Can this be merged now? 😉

DirtyF:
@jekyllbot: merge +minor

PR #358, opened 2019-09-05
remove Google+ from example snippet

coliff:
Google+ has closed down


ashmaroli:
@jekyllbot: merge +docs

PR #360, opened 2019-09-07
Fix locale specs that use the fallback locale

ljcooke:
`en_US` should not be used to test `site.lang` and `page.lang` as this is also the fallback locale.


ashmaroli:
Thanks @ljcooke 
@jekyllbot: merge +dev

PR #361, opened 2019-09-15
Adding an example of date_modified use

moxknet:
Hi! I spent all day trying to figure out how to use the "date_modified" function in jekyll-seo. Now that I understand, I would like to add an example to help others with the same problem as me.
Thanks!


DirtyF:
According to YAML spec: here are the different timestamps you can use: 

```
canonical: 2001-12-15T02:59:43.1Z
iso8601: 2001-12-14t21:59:43.10-05:00
spaced: 2001-12-14 21:59:43.10 -5
date: 2002-12-14
```

PR #367, opened 2019-10-31
Ensure a single leading `@` for twitter usernames

samtrion:
This prevents destroying the template due to the incorrect use of quotes.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -64,10 +64,10 @@
 {% endif %}
 
 {% if site.twitter %}
-  <meta name="twitter:site" content="{{ site.twitter.username | replace:'@','' | prepend:'@' }}" />
+  <meta name="twitter:site" content="{{ site.twitter.username | remove:'@' | prepend:'@' }}" />
 
   {% if seo_tag.author.twitter %}
-    <meta name="twitter:creator" content="{{ seo_tag.author.twitter | replace:'@','' | prepend:'@'  }}" />
+    <meta name="twitter:creator" content="{{ seo_tag.author.twitter | remove:'@' | prepend:'@'  }}" />
   {% endif %}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
### *Did you know?*

Liquid also has a [`remove` filter](https://github.com/Shopify/liquid/blob/57c9cf64ebc777fe5e92d4408d31a911f087eeb4/lib/liquid/standardfilters.rb#L269-L272) that seems to be more appropriate than `replace` in this use-case.

samtrion:
Done

ashmaroli:
Thanks for updating. Do you by any chance know if `@` is allowed after the first char? (e.g. `@zombie@`)

samtrion:
https://help.twitter.com/en/managing-your-account/twitter-username-rules

It's not allowed

lib/template.html
```diff
@@ -64,10 +64,10 @@
 {% endif %}
 
 {% if site.twitter %}
-  <meta name="twitter:site" content="@{{ site.twitter.username | replace:"@","" }}" />
+  <meta name="twitter:site" content="{{ site.twitter.username | remove:'@' | prepend:'@' }}" />
```
ashmaroli:
```suggestion
  <meta name="twitter:site" content="@{{ site.twitter.username | remove:'@' }}" />
```

lib/template.html
```diff
@@ -64,10 +64,10 @@
...
 
   {% if seo_tag.author.twitter %}
-    <meta name="twitter:creator" content="@{{ seo_tag.author.twitter }}" />
+    <meta name="twitter:creator" content="{{ seo_tag.author.twitter | remove:'@' | prepend:'@'  }}" />
```
ashmaroli:
```suggestion
    <meta name="twitter:creator" content="@{{ seo_tag.author.twitter | remove:'@' }}" />
```

samtrion:
Done

parkr:
A test could be helpful so we don't regress in the future!

DirtyF:
@jekyllbot: merge +minor

PR #376, opened 2020-01-31
Reduce allocations of instance-agnostic objects

ashmaroli:
So that they're not initialized for every instance of the class.
`fallback_data` is allowed to be mutated via `#[]=` therefore it cannot be stashed in a constant but is instead optimized per instance by memoizing it.


ashmaroli:
I just realized `fallback_data` is not considered as an immutable object by Jekyll Core..

Jekyll core has the following code:
```ruby
# lib/jekyll/drops/drop.rb

def []=(key, val)
  setter = "#{key}="
  if respond_to?(setter)
    public_send(setter, val)
  elsif respond_to?(key.to_s)
    if self.class.mutable?
      mutations[key] = val
    else
      raise Errors::DropMutationException, "Key #{key} cannot be set in the drop."
    end
  else
    fallback_data[key] = val
  end
end
```
The last `else` branch *mutates* `fallback_data`.
So should I revert stashing `:fallback_data` hash in a frozen constant..?

parkr:
Good catch. I’m really *surprised* that the fallback data is mutated there. The whole point of the mutations array is to hold all changes — the fallback data shouldn’t be modified. What is the commit data for that change? Something just slip past us, or did I design it this way? If so, I can’t remember why I did it that way at all.

ashmaroli:
> What is the commit data for that change? Something just slip past us, or did I design it this way?

The change was introduced in Dec 2015 via https://github.com/jekyll/jekyll/commit/3fa8af2a18b1f79dbb9c6af24894bdf5b7122d76. Looks like it was a conscious decision.
So, I'll just revert changing `fallback_data` in this plugin's drop.

ashmaroli:
Requesting a re-review..

DirtyF:
@jelkyll: merge +fix

PR #382, opened 2020-03-19
style: align with latest jekyll-rubocop

DirtyF:
Fixes some minor styles issues reported by rubocop 0.80.1 (current version used in rubocop-jekyll plugin)


lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -76,7 +76,7 @@ def main_entity
       alias_method :mainEntityOfPage, :main_entity
       private :main_entity
 
-      def to_json
+      def to_json(*_args)
```
ashmaroli: `*args` initializes an Array instance on every call.
Since `args` is not being consecutively used, this is unnecessary allocation just to appease RuboCop.
The better alternative would be to just disable the cop for this one method.

ashmaroli:
Thanks @DirtyF 
@jekyllbot: merge +dev

PR #385, opened 2020-04-13
chore(ci): bump Ruby versions

DirtyF:
Ruby 2.4 is now EOL


DirtyF:
@jekyllbot: merge +minor

PR #388, opened 2020-05-03
Render og:locale meta only when defined explicitly

### LINKED ISSUE #375: lang != locale
From https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/usage.md:

> * `lang` - The locale these tags are marked up in. Of the format `language_TERRITORY`. Default is `en_US`.

Whatever `lang` is set to seems to be used both for the `lang` attribute in the `html` tag as well as for the `og:locale` property. 
HTML's `lang` should adhere to [ISO 639-1](https://www.w3schools.com/tags/ref_language_codes.asp) (that is, language only) but `og:locale` should be on the [`language_TERRITORY` format](https://ogp.me/).

Suggestion: Add `locale` config setting and use that for `og:locale`. Use `lang` setting only for HTML's `lang`.

ashmaroli:
Render `og:locale` meta attribute only if `{{ page.locale }}` has been defined.
This change is so that `<html lang="{{ page.lang }}">` can be set to valid language (without territory) information.
  
Closes #375

### DIFF
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -164,6 +164,12 @@ module Jekyll
         @page_lang ||= page["lang"] || site["lang"] || "en_US"
       end
 
+      def page_locale
+        return @page_locale if defined?(@page_locale)
+
+        @page_locale = page["locale"] && page["locale"].tr("-", "_")
+      end
+
       def canonical_url
         @canonical_url ||= begin
           if page["canonical_url"].to_s.empty?
--- a/lib/jekyll-seo-tag/drop.rb
+++ b/lib/jekyll-seo-tag/drop.rb
@@ -165,9 +165,7 @@ module Jekyll
       end
 
       def page_locale
-        return @page_locale if defined?(@page_locale)
-
-        @page_locale = page["locale"] && page["locale"].tr("-", "_")
+        @page_locale ||= (page["locale"] || site["locale"] || page_lang).tr("-", "_")
       end
 
       def canonical_url
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -60,7 +60,7 @@ webmaster_verifications:
   baidu: 1234
 ```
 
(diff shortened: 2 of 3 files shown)


Sti2nd:
Nice, you're fast @ashmaroli ! 👏

lib/jekyll-seo-tag/drop.rb
```diff
@@ -164,6 +164,12 @@ def page_lang
         @page_lang ||= page["lang"] || site["lang"] || "en_US"
       end
 
+      def page_locale
+        return @page_locale if defined?(@page_locale)
+
+        @page_locale = page["locale"] && page["locale"].tr("-", "_")
```
shushugah, line 170: You can remove `return @page_locale if defined?(@page_locale)` and use the one liner below instead: 

`@page_locale ||= page["locale"] && page["locale"].tr("-", "_")`

ashmaroli: There is a small difference though.
With the use of `||=`, `page["locale"]` is evaluated on every call to the method if `page["locale"]` is `nil` or `false`.
However, the `defined?` check runs once per instance.

shushugah:
Would this enable/allow a default `locale` in the `_config.yml` file? E.g in case of [techworkersberlin.com](https://techworkersberlin.com), the default should be `en_DE` and occasionally on specific pages it's `de_DE`

shushugah:
And tests need to be updated to remove `<meta property="og:locale" content="en_US" />` in places where it's not explicitly set. The other option is to keep default of `en_US` but still use `page_locale` as the method/variable to set it.

DirtyF (commented):
Going from lang to locale looks like a breaking change

ashmaroli:
Yes, it **is a change in behavior**, but not exactly *breaking*.
`locale` setting takes preference over `lang`.
```ruby
def page_locale
  (page["locale"] || site["locale"] || page["lang"] || site["lang"] || "en_US").tr("-", "_")
end
```

DirtyF:
@jekyll: merge +minor

Issue #389, opened 2020-05-14
og:image and og:type not always set


LorenzBung:
If I see this correctly, the `og:image` is set to `image.path` (in [this line](https://github.com/jekyll/jekyll-seo-tag/blob/6135021410981d2d76878568e9c21cf363882bfe/lib/template.html#L33)). But if only `image: /url/to/image` is specified, the tag is not set because `image.path` is not present, right? Shouldn't there be the following:

```liquid
{% if seo_tag.image %}
    {% if seo_tag.image.path %}
        <meta property="og:image" content="{{ seo_tag.image.path }}" />
    {% else %}
        <meta property="og:image" content="{{ seo_tag.image }}" />
    {% endif %}
    ...
{% endif %}
```

Also, `og:type` is not set anywhere when setting `seo.type`, [as specified in the advanced usage guide](https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/advanced-usage.md#customizing-json-ld-output). Am I doing something wrong here? I tried setting `seo.type` in the `_config.yml` and also setting it in the front matter of pages and posts.

ashmaroli:
Alternatively, we could also use the `default` filter:
```xml
{% if seo_tag.image %}
    {% if seo_tag.image.path %}
       <meta property="og:image" content="{{ seo_tag.image.path | default: seo_tag.image }}" />
    {% endif %}
    ...
{% endif %}
```
But ultimately with either of the route, the problem is that if `seo_tag.image` is a Hash that has values set only for `height` and `width`, then the value for the meta tag's attribute `content` would end up being incorrect.

LorenzBung:
I think this should suffice as workaround:

```liquid
{% if seo_tag.image %}
    {% if seo_tag.image.path %}
       <meta property="og:image" content="{{ seo_tag.image.path }}" />
    {% else %}
        {% if seo_image.width or seo_image.height %}
            /* width and height are set, but image is an object -> do nothing */
        {% else %}
            /* Put the checks for width and height here */
        {% endif %}
    {% endif %}
{% endif %}
```

Ugly solution, but I think [Liquid has no `not`, right](https://shopify.github.io/liquid/basics/operators/)?

LorenzBung:
Even better solution, using {the `first` filter of liquid](https://shopify.dev/docs/themes/liquid/reference/filters/array-filters#first):

```liquid
{% if seo_tag.image.first %}
    /* image is object */
    {% if seo_tag.image.path %}
[...]

PR #390, opened 2020-05-14
Better checks for seo_tag.image

LorenzBung:
References #389

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -29,20 +29,16 @@
   <meta property="og:site_name" content="{{ seo_tag.site_title }}" />
 {% endif %}
 
-{% if seo_tag.image.first %}
-  {% if seo_tag.image.path %}
-    <meta property="og:image" content="{{ seo_tag.image.path }}" />
-    {% if seo_tag.image.width %}
-      <meta property="og:image:width" content="{{ seo_tag.image.width }}" />
-    {% endif %}
-    {% if seo_tag.image.height %}
-      <meta property="og:image:height" content="{{ seo_tag.image.height }}" />
-    {% endif %}
+{% if seo_tag.image.first and seo_tag.image.path %}
+  <meta property="og:image" content="{{ seo_tag.image.path }}" />
+  {% if seo_tag.image.width %}
+    <meta property="og:image:width" content="{{ seo_tag.image.width }}" />
   {% endif %}
-{% else %}
-  {% if seo_tag.image %}
-    <meta property="og:image" content="{{ seo_tag.image }}" />
+  {% if seo_tag.image.height %}
+    <meta property="og:image:height" content="{{ seo_tag.image.height }}" />
   {% endif %}
+{% elsif seo_tag.image %}
+  <meta property="og:image" content="{{ seo_tag.image }}" />
 {% endif %}
 
 {% if page.date %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -33,10 +33,10 @@
   {% if seo_tag.image.path %}
     <meta property="og:image" content="{{ seo_tag.image.path }}" />
     {% if seo_tag.image.width %}
-      <meta property="og.image:width" content="{{ seo_tag.image.width }}" />
+      <meta property="og:image:width" content="{{ seo_tag.image.width }}" />
     {% endif %}
     {% if seo_tag.image.height %}
-      <meta property="og.image:height" content="{{ seo_tag.image.height }}" />
+      <meta property="og:image:height" content="{{ seo_tag.image.height }}" />
     {% endif %}
   {% endif %}
 {% else %}
--- a/lib/template.html
+++ b/lib/template.html
@@ -29,13 +29,19 @@
   <meta property="og:site_name" content="{{ seo_tag.site_title }}" />
 {% endif %}
 
-{% if seo_tag.image %}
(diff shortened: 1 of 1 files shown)


ashmaroli:
Have you tested this, manually at least?
`first` as a filter is not invoked with the dot operator (`.`) but via the pipe operator (`|`)

LorenzBung:
Yes I have, from [the documentation](https://shopify.dev/docs/themes/liquid/reference/filters/array-filters#first):

> You can use last with dot notation when you need to use the filter inside a tag.

ashmaroli:
Okay I concede. But do note that the Liquid you're referring to is for Shopify themes, Jekyll uses the version as referenced by [this documentation](https://shopify.github.io/liquid/filters/first/).

LorenzBung:
You're right about that, I'll keep that in mind

Sena (approved):

lib/template.html
```diff
@@ -29,13 +29,19 @@
...
+    {% if seo_tag.image.width %}
+      <meta property="og:image:width" content="{{ seo_tag.image.width }}" />
+    {% endif %}
+    {% if seo_tag.image.height %}
+      <meta property="og:image:height" content="{{ seo_tag.image.height }}" />
+    {% endif %}
   {% endif %}
-  {% if seo_tag.image.width %}
-    <meta property="og:image:width" content="{{ seo_tag.image.width }}" />
+{% else %}
```
shushugah, line 42:
```
{% elsif seo_tag.image %}
```

lib/template.html
```diff
@@ -29,13 +29,19 @@
   <meta property="og:site_name" content="{{ seo_tag.site_title }}" />
 {% endif %}
 
-{% if seo_tag.image %}
-  <meta property="og:image" content="{{ seo_tag.image.path }}" />
-  {% if seo_tag.image.height %}
-    <meta property="og:image:height" content="{{ seo_tag.image.height }}" />
+{% if seo_tag.image.first %}
```
shushugah, line 32: `{% if seo_tag.image.first and seo_tag.image.path %}`

shushugah (commented):
suggestion to combine nested if/else conditions

ashmaroli:
@LorenzBung This needs tests.
It is not clear what `seo_tag.image.first` returns nor the purpose of checking `seo_tag.image.first` when it is not subsequently used.

Also, I see a lot of `seo_tag.image` calls now. I think you could stash that in a variable and reference the variable thereafter..
```liquid
{% if seo_tag.image %}
  {% assign abcd = seo_tag.image %}
  <meta property="og:image" content="{{ abcd.path }}" />
  {% if abcd.width %}
    <meta property="og:image:width" content="{{ abcd.width }}" />
  {% endif %}
  ...
{% endif %}
```

LorenzBung:
> It is not clear what `seo_tag.image.first` returns nor the purpose of checking `seo_tag.image.first` when it is not subsequently used.

The value it returns is irrelevant here. I just want to check whether `seo_tag.image` contains a string (=> `seo_tag.image` is the image URL itself) or if it is a dictionary, (=> `seo_tag.image.path` contains the URL).
In the latter case, `seo_tag.image.first` will contain some value which is why I check it.

> Also, I see a lot of seo_tag.image calls now. I think you could stash that in a variable and reference the variable thereafter..

Yeah I could but honestly I found it quite readable the way it currently is - if you want I can add some variable though.

ashmaroli:
> The value it returns is irrelevant here..

I took a peak under the hood and the tests:
https://github.com/jekyll/jekyll-seo-tag/blob/22a160cb32681a4349db7105626ad6aa0b1b2954/lib/jekyll-seo-tag/drop.rb#L106-L111
https://github.com/jekyll/jekyll-seo-tag/blob/22a160cb32681a4349db7105626ad6aa0b1b2954/spec/jekyll_seo_tag/image_drop_spec.rb#L17-L22

So basically, if a `page` has `image: foo.png` in the front matter, then `seo_tag.image` results in a truthy.
Therefore, I think this pull request is unnecessary entirely..

LorenzBung:
> So basically, if a `page` has `image: foo.png` in the front matter, then `seo_tag.image` results in a truthy.
> Therefore, I think this pull request is unnecessary entirely..

The reason for this PR is that `image: foo.png` would not be added as meta tag correctly because the liquid template expects `image.path: foo.png`.

ashmaroli:
> The reason for this PR is that `image: foo.png` would not be added as meta tag correctly

@LorenzBung Can you demonstrate this with a test-site?
Since this plugin is whitelisted on GitHub Pages, you will be able to demo easily as well..
Just create a new site locally with `jekyll new ...` and add `image: foo.png` to `about.md` and `image.path: foo.png` in a new page named `test.md`.
Push the site onto GitHub and enable GitHub Pages builds on the repo.

LorenzBung:
Hm, weird, all my tries to reproduce this bug were not successful. The only thing I observed was `image.path: foo.png` not resulting in any output, while `image: foo.png` and `image:\n    path: foo.png` did.

Sorry for wasting your time.

ashmaroli:
> The only thing I observed was `image.path: foo.png` not resulting in any output, while `image: foo.png` and `image:\n path: foo.png` did.

The reason is `image.path` is not a valid data writer. It is a reference to the following YAML front matter:
```
---
image:
  path: foo.png
---
```

PR #391, opened 2020-05-17
Set the default og:type to 'website'

CodeSandwich:
If the page doesn't have a `date`, `og:type` is not set. This PR attempts to fix that.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -42,6 +42,8 @@
 {% if page.date %}
   <meta property="og:type" content="article" />
   <meta property="article:published_time" content="{{ page.date | date_to_xmlschema }}" />
+{% else %}
+  <meta property="og:type" content="website" />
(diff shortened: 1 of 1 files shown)


LorenzBung:
Relevant update for #389

neilconway:
Any updates on this? Would be nice to have this fix -- `og:type` is mandatory.

LorenzBung:
> Any updates on this? Would be nice to have this fix -- `og:type` is mandatory.

The way I see it, this PR should be ready to be merged as it fixes `og:type` being unassigned.
The type itself might be changed in a separate PR, but as far as I understand it, this one is only there to set a default value.

PardhuMadipalli (approved):

PardhuMadipalli:
Why is nobody reviewing this PR? This is an important fix.

PR #393, opened 2020-05-21
Disable other title usage along wit title tag

### LINKED ISSUE #373: i18n: Plugin does not work with jekyll-multiple-languages-plugin
Been knee deep with a i18n project, and this plugin does not work well with page title tags (maybe other elements too, haven't investigated.) 

The `jekyll-multiple-languages-plugin` [1] works by changing the title frontmatter to refer to separate language files placed in `i18n/language-name/posts/blog.md` where the actual translated text lives. As such:

```
---
title: posts.blog.title
subtitle: posts.blog.subtitle
---

{% tf posts/blog.md %}
```

The end result is translated files, but bad seo, as titles are being outputted as a literal `posts.blog.title`.

Now I am willing to look into this plugin and figure out how to do it and then do a PR, but I am not the most well-versed in Ruby. Any pointers or help would be highly appreciated.

---

[1] [Jekyll Multiple Languages Plugin](https://github.com/kurtsson/jekyll-multiple-languages-plugin)

trasher:
Hi,

I do not know why this disable title option has been added; I'm trying to fix #373.

Problem is my pages titles looks like `pages.home` outputted verbatim in title, and 2 other meta tags. DIsabling title did the trick only for the `<title>` element, my goal is to replace on my side missing metas.

I'm not a ruby developper (been using jekyll for a few days only), and I've not been able to find a solution to override `template.html`; which would have been more interesting perhaps.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -12,6 +12,8 @@ If for some reason, you don't want the plugin to output `<title>` tags on each p
 ```
 <!-- {% endraw %} -->
 
+This will remove output for `<title>`, `<meta property="og:title"/>` and `<meta property="twitter:title"/>`; you can define each of them on your side if needed.
+
 ### Author information
 
 Author information is used to propagate the `creator` field of Twitter summary cards. This should be an author-specific, not site-wide Twitter handle (the site-wide username be stored as `site.twitter.username`).
--- a/lib/template.html
+++ b/lib/template.html
@@ -5,7 +5,7 @@
 
 <meta name="generator" content="Jekyll v{{ jekyll.version }}" />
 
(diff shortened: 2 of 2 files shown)


trasher:
Extra note: I had another issue: the `application/ld+json` contains also the page title, so it gets the untranslated strings... I've fixed using a replace in the `template.html` file, did not found another way.

Sena (approved):

Sena (approved):

centeno (approved):

centeno (approved):

centeno:
@ashmaroli can you help us to approve this PR ?

trasher:
It's been several years without any answer from upstream; I close.

PR #403, opened 2020-06-16
Rename Google webmaster tools to Google Search Console 

ptcodes:
- Google webmaster tools no longer exists as it has been renamed to Google Search Console
- Fixed several links 
- Changed front-matter to front matter to be consistent across the docs


DirtyF:
@jekyll: merge +docs

Issue #396, opened 2020-06-05
Customizing template.html


racerxdl:
The issue https://github.com/jekyll/jekyll-seo-tag/issues/355 was closed by the bot. But it is useful to customize the template.html.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

racerxdl:
up

pudiva:
bump. I'd really like to customise some logic there:
* add support for __og:video__
* automatically generate image and video urls based on id stored on front-matter
* reuse social links that are stored in another config key

pudiva:
my hope is that such feature would get into gh-pages, for maximum ~laziness~ convenience

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

racerxdl:
bump

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

pudiva:
bump

jekyllbot:
[...]

Issue #355, opened 2019-08-02
Customizing template.html


regpaq:
First, thanks to all for making this awesome plugin. 

Simple question: can I customize the template.html file? For example, is it possible to make simple changes like so:

**Before**
`<meta property="og:image" content="{{ seo_tag.image.path }}" />`

**After**
`<meta property="og:image" content="{{site.url}}/assets/images/{{ page.image }}" />`

Reason I ask is this works perfectly in my local environment and looks correct on github in my _site folder, but the live site (deployed with netlify) is using the before code. Any idea what's causing this?

Thanks in advance

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

racerxdl:
Up

PR #404, opened 2020-06-22
option to skip meta tag generator

Sena:
With this option we allow the dev choose if they wanna show this meta tag or not

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -3,7 +3,9 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
+{% unless site.skip_generator %}
 <meta name="generator" content="Jekyll v{{ jekyll.version }}" />
+{% endunless %}
 
 {% if seo_tag.page_title %}
   <meta property="og:title" content="{{ seo_tag.page_title }}" />


rjd22:
:+1: I would also like this option.

Sena:
@DirtyF can you help this PR

mrdaliri:
I like to have this option. But I prefer something more general which allows you to disable any of generated tags. Like `{% seo generator=false %}

davorpa:
Ref #355 #396 

Skip `generator` meta tag is a good idea (or at least doesn't show version).

**It's well-known to be used to know if site has any vulnerability.**

The less information is given about frameworks, application servers, libraries ... the better.

lib/template.html
```diff
@@ -3,7 +3,9 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
+{% unless site.skip_generator %}
```
davorpa, line 6: Could you change it by 

```
{% unless seo_tag.skip_generator %}
```

It's more general

davorpa (changes requested):
Ref #355 #396 

Skip `generator` meta tag is a good idea (or at least doesn't show version).

**It's well-known to be used to know if site has any vulnerability.**

The less information is given about frameworks, application servers, libraries ... the better.

davorpa:
Unit tests must be also provided to check right working

https://github.com/jekyll/jekyll-seo-tag/blob/93c664bce07aa290e141f15763dbfefe5de2744b/spec/jekyll_seo_tag_integration_spec.rb#L22-L31

https://github.com/jekyll/jekyll-seo-tag/blob/93c664bce07aa290e141f15763dbfefe5de2744b/spec/jekyll_seo_tag_integration_spec.rb#L299-L313

mattr-:
Should we simply leave out the generator tag completely?

rafaelbiriba:
Any solution on this?

danger89:
Follow-up, PR: https://github.com/jekyll/jekyll-seo-tag/pull/481

Issue #394, opened 2020-05-29
Jekyll 4.1.0 breaks Travis tests


emmasax4:
Using Jekyll version 4.1.0 (which was released the other day) breaks Travis tests. The `.travis.yml` file should lockdown Jekyll to `~> 4.0.0` or fix the tests: https://github.com/jekyll/jekyll-seo-tag/blob/master/.travis.yml#L18.

Long story short, I was fiddling around with a fork, and found that the later version breaks: https://travis-ci.com/github/emma-sax4/jekyll-seo-tag/builds/168817663. But locking down Travis succeeds again: https://travis-ci.com/github/emma-sax4/jekyll-seo-tag/builds/168820682.

ashmaroli:
@emmasax4 Jekyll 4.1.1 has been shipped. Can you verify if this is still an issue?
Thanks.

emmasax4:
Yep, the issue is resolved now. I'll make [a pull request](https://github.com/jekyll/jekyll-seo-tag/pull/405) to lock down the version of Jekyll that Travis uses.

PR #405, opened 2020-06-24
Force higher version of Travis to use higher version of Jekyll

emmasax4:
I created [this issue](https://github.com/jekyll/jekyll-seo-tag/issues/394) a while ago because I learned that the newest version of Jekyll had broken the tests on this repository. Now Jekyll has released a new version, but we still should force Travis to test with Jekyll versions that are at least 4.1.1 (not 4.1.0).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,4 +15,4 @@ env:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   matrix:
     - JEKYLL_VERSION="~> 3.8"
-    - JEKYLL_VERSION="~> 4.0"
+    - JEKYLL_VERSION="~> 4.1.1"


emmasax4:
@ashmaroli This is the follow up pull request the issue linked above. Can you give it a review? Thank you!

DirtyF:
That won't solve the issue as we don't schedule plugin builds before a release. 
Plus we want to keep the [pessimistic operator ](https://thoughtbot.com/blog/rubys-pessimistic-operator) to minor versions  `~> 4.1` to avoid having to change it until next major version.

ashmaroli:
@emmasax4 FYI, there's no need to explicitly test v4.1.1 and beyond.
- `~> 4.0` means accept all versions from `v4.0.0` onwards until `v5.0.0` i.e. ones in the `v4.x.x` series.
- `~> 4.1.1` would mean accept only versions from `v4.1.x` series. The future `v4.2.0` won't get tested.

Regarding this PR, there is no need for this change at all since if you restart your CI builds, it will automatically use the latest version available.

emmasax4:
Ah okay. Yeah I got that having 3 digits would limit it more, but figured it was fine for now and people could change it again later. But that's fine. Good to know that this helps it.

PR #415, opened 2020-09-21
Fix Travis builds for Jekyll 3.x

ashmaroli:
Jekyll 3.9 requires gem `kramdown-parser-gfm` to be additionally made available in the environment.


DirtyF (approved):

DirtyF:
@jekyll: merge +fix

PR #417, opened 2020-10-02
Mutate hash literals instead of duplicating them

ashmaroli:
Prefer `Hash#merge!` over `Hash#merge` for temporary Hashes.
*At a micro-level, the mutating method [is faster](https://github.com/JuanitoFatas/fast-ruby#hashmerge-vs-hashmerge-code) as well.*

Additionally, stash `page["image"]` in a local variable in `ImageDrop` so as to reduce calls to `Jekyll::Drops::DocumentDrop#[]`


parkr (approved):
If tests pass, 👍

DirtyF (approved):
What Parkr said 😄

DirtyF:
@jekyll: merge + dev

PR #427, opened 2021-01-06
Allow to set type for author

api-walker:
Referring to https://schema.org/author the values Organization and Person are supported

### DIFF
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -21,7 +21,9 @@ module Jekyll
       private :logo
 
       VALID_ENTITY_TYPES = %w(BlogPosting CreativeWork).freeze
+      VALID_AUTHOR_TYPES = %w(Organization Person).freeze
       private_constant :VALID_ENTITY_TYPES
+      private_constant :VALID_AUTHOR_TYPES
 
       # page_drop should be an instance of Jekyll::SeoTag::Drop
       def initialize(page_drop)
@@ -36,10 +38,13 @@ module Jekyll
       end
 
       def author
+        author_type = page_drop.author["type"]
+
         return unless page_drop.author["name"]
+        return nil if author_type && !VALID_AUTHOR_TYPES.include?(author_type)
 
         {
-          "@type" => "Person",
+          "@type" => author_type || "Person",
           "name"  => page_drop.author["name"],
         }
       end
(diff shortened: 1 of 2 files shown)


DarkWiiPlayer:
This addresses one possible use case of my recent pull request but makes it a lot less hacky. It'd be nice if they could both be merged to make this plugin viable for sites other than blogs. 🎆 🎊

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -21,7 +21,9 @@ class JSONLDDrop < Jekyll::Drops::Drop
       private :logo
 
       VALID_ENTITY_TYPES = %w(BlogPosting CreativeWork).freeze
+      VALID_AUTHOR_TYPES = %w(Organization Person).freeze
       private_constant :VALID_ENTITY_TYPES
+      private_constant :VALID_AUTHOR_TYPES
```
ashmaroli, line 26: Please add this symbol to the preceding line:
```ruby
private_constant :VALID_ENTITY_TYPES, :VALID_AUTHOR_TYPES
```

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -36,10 +38,13 @@ def fallback_data
       end
 
       def author
+        author_type = page_drop.author["type"]
```
ashmaroli, line 41: I prefer having this variable assigned after the check for `author["name"]` so that the variable isn't assigned unnecessarily.

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -36,10 +38,13 @@ def fallback_data
...
+
         return unless page_drop.author["name"]
+        return nil if author_type && !VALID_AUTHOR_TYPES.include?(author_type)
```
ashmaroli, line 44: let us be consistent with the preceding line and return `nil` implicitly:
```suggestion
        return if author_type && !VALID_AUTHOR_TYPES.include?(author_type)
```

ashmaroli (commented):
Some minor changes needed..

ashmaroli (approved):

parkr (approved):
This LGTM

ashmaroli:
Thank you @api-walker 
@jekyllbot: merge +minor

PR #428, opened 2021-01-06
Add option for overriding structured data attributes

DarkWiiPlayer:
In the current implementation, only certain attributes can be set for
the structured data. This makes it impossible to use certain types like
"SoftwareApplication", which need unsupported attributes for google to
accept them for rich results. 😖

This change allows the user to override attributes in the structured
data tree by recursively merging in attributes from the
custom_structured_data subtree.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -78,6 +78,7 @@ The following options can be set for any particular page. While the default opti
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
   * `date_modified` - Manually specify the `dateModified` field in the JSON-LD output to override Jekyll's own `dateModified`.
   This field will take **first priority** for the `dateModified` JSON-LD output. This is useful when the file timestamp does not match the true time that the content was modified. A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
+  * `custom_structured_data` - Override single values in the structured data tree. Merging is recursive so nested attributes can be overridden without overriding the whole tree.
 
 ### Customizing image output
 
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -79,8 +79,22 @@ module Jekyll
       alias_method :mainEntityOfPage, :main_entity
       private :main_entity
 
+      private def deep_merge(this_hash, other_hash)
+        this_hash.merge(other_hash) do |_key, this_val, other_val|
+          if this_val.is_a?(Hash) && other_val.is_a?(Hash)
+            deep_merge(this_val, other_val)
+          else
+            other_val
+          end
+        end
+      end
+
       def to_json
-        to_h.reject { |_k, v| v.nil? }.to_json
+        drop_hash = to_h
+        if page_drop.custom_structured_data.is_a? Hash
+          drop_hash = deep_merge(drop_hash, page_drop.custom_structured_data)
+        end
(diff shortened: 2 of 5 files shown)


lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -79,8 +79,22 @@ def main_entity
       alias_method :mainEntityOfPage, :main_entity
       private :main_entity
 
+      private def deep_merge(this_hash, other_hash)
```
ashmaroli, line 82: Two issues here:
- Jekyll Core already has a method for similar task: `Jekyll::Utils.deep_merge_hashes`. Why re-invent the wheel?
- This code base doesn't use inline private calls for explicit method definitions. Private methods are defined at the bottom of a class under the `private` keyword.

ashmaroli (changes requested):

lib/jekyll-seo-tag/json_ld_drop.rb
DarkWiiPlayer: Ah, I see. I wasn't aware of that builtin function, so I've now removed the custom one and changed it to use Jekylls version instead.

api-walker:
I have an use case for this pull request too.

docs/advanced-usage.md
```diff
@@ -78,6 +78,7 @@ The following options can be set for any particular page. While the default opti
   * `links` - An array of other URLs that represent the same thing that this page represents. For instance, Jane's bio page might include links to Jane's GitHub and Twitter profiles.
   * `date_modified` - Manually specify the `dateModified` field in the JSON-LD output to override Jekyll's own `dateModified`.
   This field will take **first priority** for the `dateModified` JSON-LD output. This is useful when the file timestamp does not match the true time that the content was modified. A user may also install [Last Modified At](https://github.com/gjtorikian/jekyll-last-modified-at) which will offer an alternative way of providing for the `dateModified` field.
+  * `custom_structured_data` - Override single values in the structured data tree. Merging is recursive so nested attributes can be overridden without overriding the whole tree.
```
ashmaroli, line 81: Could you add an example for this use-case i.e. *overriding nested attributes*.

api-walker:
Is there any news here?

ashmaroli (approved):

DirtyF (approved):

DarkWiiPlayer:
Is there any further work required on my side before this can be merged?

api-walker:
Is there any progress here yet?

DarkWiiPlayer:
@ashmaroli @DirtyF Is there anything still preventing this PR from being merged?

DirtyF:
@DarkWiiPlayer I was hoping @benbalter or someone from GitHub would give formal approval, as this is a white-listed plugin on GitHub Pages and we don't want to introduce any security issue.

DarkWiiPlayer:
> @DarkWiiPlayer I was hoping @benbalter or someone from GitHub would give formal approval, as this is a white-listed plugin on GitHub Pages and we don't want to introduce any security issue.

Are you still hoping?

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -84,7 +84,11 @@ def main_entity
       private :main_entity
 
       def to_json
-        to_h.reject { |_k, v| v.nil? }.to_json
+        drop_hash = to_h
+        if page_drop.custom_structured_data.is_a? Hash
+          drop_hash = Utils.deep_merge_hashes(drop_hash, page_drop.custom_structured_data)
+        end
+        drop_hash.reject { |_k, v| v.nil? }.to_json
```
ashmaroli, line 91: @DarkWiiPlayer This plugin now requires at least Ruby 2.4.0.
So, could you bring the branch up to date and use `drop_hash.compact` here?

Issue #424, opened 2020-12-12
Logo Rich Text Results Type in Unsupported Format


benjithaimmortal:
I think Google has updated their Rich Results Text matching criteria for Logos

I'm testing this on my [personal site](https://www.monjibram.com/) as an example: it seems that [Google's Rich Text Results for a logo](https://developers.google.com/search/docs/data-types/logo) are only filling when the LD+JSON is set to `@type: Organization`, with `logo: [img_url]` and `url: [baseurl]`. `jekyll-seo-tag` is outputting them inside of `"Publisher":` which is not recognized by the test.

https://search.google.com/test/rich-results?url=https%3A%2F%2Fwww.monjibram.com%2F&user_agent=1

If someone can confirm that I'm implementing this correctly in _config.yml, I think this needs a change:
```
logo: assets/images/Benji.jpg
type: Organization
```

I've identified the potential area in the source code: `lib/jekyll-seo-tag/json_ld_drop.rb` lines 23 and 57.

### Add Type "Organization" to valid types
Line 23: `VALID_ENTITY_TYPES = %w(BlogPosting CreativeWork).freeze`
Change to: `VALID_ENTITY_TYPES = %w(BlogPosting CreativeWork Organization).freeze`

### Update this 
`publisher` Method on Line 57:
```
      def publisher
        return unless logo

        output = {
          "@type" => "Organization",
          "logo"  => {
            "@type" => "ImageObject",
            "url"   => logo,
          },
        }
        output["name"] = page_drop.author.name if page_drop.author.name
        output
      end
```
Change to: 
```
      def publisher
        return unless logo && (page_drop.type == 'Organization')
        output["logo"] = logo
        output["name"] = page_drop.author.name if page_drop.author.name
        output
      end
```

I haven't written Ruby in many moons, so please forgive any errors!

benjithaimmortal:
With a bit more research, I came across this as an alt `output` format for the `publisher` method:
```
 output = {"@graph": [{
    "@type" => "Organization",
    "logo": {
      "@type" => "ImageObject",
      "url" => url
    },
    "name": page_drop.author.name,
    "url": page_drop.canonical_url
  }]}
```

ashmaroli:
@benjithaimmortal You're free to formalize your suggestions by submitting a pull request with your change and adding / updating tests as necessary.

benjithaimmortal:
[...]

PR #429, opened 2021-01-13
This way to a passing Rich Results logo test

benjithaimmortal:
#### A forward
@ashmaroli Thank you for your patience as I got enough Ruby under my belt to even get this far. I started using `pry` to debug the code in-stream, and it allowed me to even test the output as it was rendered in `rspec`. This is now passing the test on Google, and passing `rspec` and `rubocop`. I don't think my code is great, but it was an attempt. My first! Let's change any/all of it in the hopes that the overall contribution helps this cool open source project.
--Benji

-------

## TL;DR

1. We need a "url" field inside of `"@type" => "Organization"`
2. We need `"@type" => "Organization"` to always be separate from `"@type" => "WebPage"`
3. We need to get it out of `"publisher"`

### We're trying to get a set `logo` to pass a Rich Results test, and there are a few problems.
Google may have changed this, but at the moment it isn't reading things inside of the "publisher" key. As I mentioned in [my open issue](https://github.com/jekyll/jekyll-seo-tag/issues/424) we need to find a way to modify the hash to remove it from `"publisher"` (which [isn't a type property](https://schema.org/publisher) on schema.org anyway).

### My proposal: let's take a look at Yoast, because they do this a lot
Yoast likes to make tons of different `@type` attributes, and house them as an array of objects inside of `@graph`. That's what I'm finding on schema.org, too. So now we want things to look like this:
```JSON
{
  "@context": "https://schema.org",
  "@graph": [
    { "@type": "Organization", },
    { "@type": "WebSite", },
    { "@type": "ImageObject", },
    { "@type": "WebPage", },
  ]
}
```

### I'd rather manipulate the hash than metaprogram
```ruby
def to_json
  # this was what happened before
  graph = to_h.reject { |_k, v| v.nil? }

  # assign publisher and remove it from the array
  # (because I don't know the meta-programming involved in instantiating it)
  publisher = graph["publisher"] || nil
  graph.delete("publisher")

  updated_graph = {
    "@context" => "https://schema.org",
    "@graph"   => [graph],
  }
  if publisher
    # .push({"something" => blah}) === << {"something" => blah}
    updated_graph["@graph"] << publisher
  end

  # .to_json for the win
  updated_graph.to_json
end
```

### Results of the change

Original Output:
```JSON
{
  "@type": "WebPage",
  "publisher": {
    "@type": "Organization",
    "logo": { "@type": "ImageObject", "url": "http://example.invalid/logo.png" }
  },
  "url": "http://example.invalid/page.html",
  "@context": "https://schema.org"
}
```

New Output (confirmed on [Rich Results Test](https://search.google.com/test/rich-results), when you change the canonical url to a valid one):
```JSON
{
  "@context": "https://schema.org",
  "@graph": [
    {
      "@type": "WebPage",
      "url": "http://example.invalid/page.html",
      "@context": "https://schema.org"
    },
    {
      "@type": "Organization",
      "url": "http://example.invalid/page.html",
      "logo": {
        "@type": "ImageObject",
        "url": "http://example.invalid/logo.png"
      }
    }
  ]
}
```

### DIFF
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -84,7 +84,25 @@ module Jekyll
       private :main_entity
 
       def to_json
-        to_h.reject { |_k, v| v.nil? }.to_json
+        # this was what happened before
+        graph = to_h.reject { |_k, v| v.nil? }
+
+        # assign publisher and remove it from the array
+        # (because I don't know the meta-programming involved in instantiating it)
+        publisher = graph["publisher"] || nil
+        graph.delete("publisher")
+
+        updated_graph = {
+          "@context" => "https://schema.org",
+          "@graph"   => [graph],
+        }
+        if publisher
+          # .push({"something" => blah}) === << {"something" => blah}
+          updated_graph["@graph"] << publisher
+        end
+
+        # .to_json for the win
+        updated_graph.to_json
       end
 
       private
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -63,6 +63,7 @@ module Jekyll
 
         output = {
           "@type" => "Organization",
+          "url"   => page_drop.canonical_url,
           "logo"  => {
             "@type" => "ImageObject",
             "url"   => logo,
--- a/spec/jekyll_seo_tag_integration_spec.rb
+++ b/spec/jekyll_seo_tag_integration_spec.rb
@@ -236,7 +236,7 @@ RSpec.describe Jekyll::SeoTag do
       let(:site) { make_site("logo" => "/logo.png", "url" => "http://example.invalid") }
 
       it "outputs the logo" do
-        expect(json_data["publisher"]["logo"]["url"]).to eql("http://example.invalid/logo.png")
+        expect(json_data["@graph"][1]["logo"]["url"]).to eql("http://example.invalid/logo.png")
       end
     end
 
@@ -244,7 +244,7 @@ RSpec.describe Jekyll::SeoTag do
       let(:site) { make_site("logo" => "http://cdn.example.invalid/logo.png", "url" => "http://example.invalid") }
 
       it "outputs the logo" do
-        expect(json_data["publisher"]["logo"]["url"]).to eql("http://cdn.example.invalid/logo.png")
+        expect(json_data["@graph"][1]["logo"]["url"]).to eql("http://cdn.example.invalid/logo.png")
       end
     end
 
@@ -252,7 +252,7 @@ RSpec.describe Jekyll::SeoTag do
       let(:site) { make_site("logo" => "http://cdn.example.invalid/logo.png", "url" => "http://example.invalid", "author" => "Mr. Foo") }
 
       it "outputs the author" do
-        expect(json_data["publisher"]["name"]).to eql("Mr. Foo")
+        expect(json_data["@graph"][1]["name"]).to eql("Mr. Foo")
       end
     end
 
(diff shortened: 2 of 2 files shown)


Sena (approved):

DarkWiiPlayer:
If I read this correctly, this seems to be solving a different problem than #428, although there seems to be some overlap and both PRs modify the `to_json` method (though this should be trivial to merge).

@benjithaimmortal could you have a quick look at #428 and confirm whether these are two distinct problems?

benjithaimmortal:
@DarkWiiPlayer you're right! We're approaching the same point in the code but doing very different things. I think my results should precede yours, so you can modify things after the correct formatting to the JSON-LD has been applied.

`@graph` is the way mainstream tools like Yoast combine multiple Structured Data results into a single page (for example, making two `@type`s for the same page). I haven't seen another method that works quite like it. See [this blog](https://moz.com/blog/writing-structured-data-guide) heading 'Creating a node array using @graph' for a bit more on how it's working.

DarkWiiPlayer:
> I think my results should precede yours [in the code logic, that is], so you can modify things after the correct formatting to the JSON-LD has been applied.

@benjithaimmortal Yes, absolutely. However, I'm wondering how that would work with the way my PR currently works where you build a new tree of json data that gets merged into the autogenerated one because now that includes an array, so one has to throw in a numeric index.

For example, if you have a `@type: Webpage` and a `@type: Organization` and want to change the url for the Webpage, you'd have to index `@graph` with `[1]`, which seems quite arbitrary and brittle.

benjithaimmortal:
@DarkWiiPlayer yeppers. I encountered the same when I was building mine. IMO the need is a simple hash from the yaml settings, and it's overcomplicated by the Ruby magic inside of a new Jekyll instance. Perhaps it could be extracted entirely? I'd love to see something that simply builds the JSON out of `config.yml` and plops it in the document header.

benjithaimmortal:
If you're worried about the array itself, perhaps we could use `@type` as a key and just strip it from the final JSON? These feel like decisions that are above my head though.

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -84,7 +85,25 @@ def main_entity
       private :main_entity
 
       def to_json
-        to_h.reject { |_k, v| v.nil? }.to_json
+        # this was what happened before
```
ashmaroli, line 88: Please avoid *conversational* comments..
```suggestion
```

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -84,7 +85,25 @@ def main_entity
...
+        graph = to_h.reject { |_k, v| v.nil? }
+
+        # assign publisher and remove it from the array
+        # (because I don't know the meta-programming involved in instantiating it)
```
ashmaroli, line 92: Same as above.
Please avoid *conversational* comments..
```suggestion
```

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -84,7 +85,25 @@ def main_entity
...
+        publisher = graph["publisher"] || nil
+        graph.delete("publisher")
```
ashmaroli, line 94: You could merge these two lines into one:
```ruby
publisher = graph.delete("publisher")
```

lib/jekyll-seo-tag/json_ld_drop.rb
```diff
@@ -84,7 +85,25 @@ def main_entity
...
+
+        updated_graph = {
+          "@context" => "https://schema.org",
+          "@graph"   => [graph],
+        }
+        if publisher
+          # .push({"something" => blah}) === << {"something" => blah}
```
ashmaroli, line 101:
```suggestion
```

ashmaroli (changes requested):
Accessing values of `json_data[@graph]` with index numbers (since the object is an Array) seems less intuitive to me as against accessing the values with explicit keys.

Nevertheless, this proposal needs some changes.

lib/jekyll-seo-tag/json_ld_drop.rb
DarkWiiPlayer: What if `graph["publisher"]` was nil though?

benjithaimmortal:
Excellent! I'll get to it and modify this ASAP. Thank you all for your comments and help.

lib/jekyll-seo-tag/json_ld_drop.rb
benjithaimmortal: Yep, let's remove these comments!

lib/jekyll-seo-tag/json_ld_drop.rb
benjithaimmortal: I think I don't know enough about Ruby for this one. Sorry!

`publisher` should be set to the inner hash `graph['publisher']`, and not the full `graph` hash. Does the refactor do that?

PR #447, opened 2021-06-18
Add og property 'article:modified_time'.

fl3a:
Hi there,
it would be great to use `page.last_modified_at` in seotags for open graph `article:published_time` too  when it exists.
This variable is already used by _jekyll-sitemap_ and its purpose seems to be the same to me.


jdvp:
agree this would be great

mattr-:
Closing in favor of #515, now available in jekyll-seo-tag v2.9.0

PR #449, opened 2021-09-01
Fix typo in source code comment

MichaelCurrin:


lib/jekyll-seo-tag/url_helper.rb
```diff
@@ -8,8 +8,8 @@ module UrlHelper
 
       # Determines if the given string is an absolute URL
       #
-      # Returns true if an absolute URL.
```
ashmaroli, line 11: Why was this *period* removed, @MichaelCurrin?

MichaelCurrin: The two lines below and the one above do not have one. So made it consistent.

ashmaroli: Fair enough.

ashmaroli (approved):

ashmaroli:
Thanks @MichaelCurrin 
@jekyllbot: merge +dev

PR #453, opened 2021-09-20
Allow to set author url

imustafin:
Starting from August 6, 2021 [Google recommends](https://developers.google.com/search/updates) to set the `author.url` property. This property helps Google to disambiguate the correct author of the article.

This PR adds the ability to specify the author url together with the other author properties.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -68,6 +68,18 @@ There are several ways to convey this author-specific information. Author inform
   author: benbalter
   ```
 
+#### Setting author url
+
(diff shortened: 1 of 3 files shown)


imustafin:
@ashmaroli hi! I don't know if you are the correct person to ping, so sorry for disturbing.

Is there anything needed from my side to move forward with this PR?

ashmaroli:
@imustafin The proposed diff looks good to me.
Just one minor thing. Could you add a test to prove that `author["url"]` is null when not provided..?
(Appending to an existing test for `author` will suffice.)

ashmaroli:
> did you mean something like this?

Yes. That'll do.
Thanks.

PR #458, opened 2022-01-31
Sort JSON-LD data by key

### LINKED ISSUE #362: Consistent order of keys for JSON-LD
It seems almost every time I `jekyll build`, the JSON-LD output appears to have changed when it hasn't.

I use a git-based deployment and it'd be nice to not have every page on the site change for every build to make it easier to review beforehand or afterwards, as well as for size of the repository over time.

I suspect the issue is with the key order not being deterministic. Would you be open to a pull request that makes this deterministic?

cap

### LINKED ISSUE #436: Make JSON-LD output deterministic
Having the same issues reported in #362, where every rendering is appearing as a modified file and triggering unnecessary updates.  Can the output of template.html be made deterministic to support reproducible builds?

If making the output of jsonify deterministic is too difficult, perhaps #135 is a good solution to give users more control where reproducible builds are critical.

parkr:
Deterministic output is desired to improve diffability between builds.
If the output randomly changes order each time it's built, then every
page on a given site will change with every Jekyll build. Tools like
rsync and git which can diff files will always show a change even when
the content hasn't truly changed.

Fixes #362. Fixes #436.

### DIFF
--- a/spec/jekyll_seo_tag_integration_spec.rb
+++ b/spec/jekyll_seo_tag_integration_spec.rb
@@ -30,6 +30,10 @@ RSpec.describe Jekyll::SeoTag do
     expect(output).to match(%r!Jekyll v#{version}!i)
   end
 
+  it "outputs JSON sorted by key" do
+    expect(json.strip).to eql('{"@context":"https://schema.org","@type":"WebPage","url":"/page.html"}')
+  end
+
   it "outputs valid HTML" do
     site.process
     options = {
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -92,7 +92,7 @@ module Jekyll
       def to_json(state = nil)
         keys.sort.each_with_object({}) do |(key, _), result|
           v = self[key]
-          result[key] = v if v # don't write if value is nil
+          result[key] = v unless v.nil?
         end.to_json(state)
       end
 
--- a/lib/jekyll-seo-tag/json_ld_drop.rb
+++ b/lib/jekyll-seo-tag/json_ld_drop.rb
@@ -87,8 +87,13 @@ module Jekyll
       alias_method :mainEntityOfPage, :main_entity
       private :main_entity
 
-      def to_json
-        to_h.compact.to_json
+      # Returns a JSON-encoded object containing the JSON-LD data.
+      # Keys are sorted.
+      def to_json(state = nil)
+        keys.sort.each_with_object({}) do |(key, _), result|
+          v = self[key]
+          result[key] = v if v # don't write if value is nil
+        end.to_json(state)
       end
 
       private


lib/jekyll-seo-tag/drop.rb
```diff
@@ -103,6 +103,11 @@ def json_ld
         @json_ld ||= JSONLDDrop.new(self)
       end
 
+      # JSON-LD output, sorted by the key, as JSON.
+      def json_ld_sorted_json
+        JSON.parse(json_ld.to_json).sort_by(&:first).to_h.to_json
```
ashmaroli, line 108: Parsing a JSON string just to sort the Hash object seems sub-optimal to me.
How about sorting the hash at source directly?

Also, `json_ld_sorted_json` sounds repetitive.
How about simply `sorted_json_ld`?

```ruby
def sorted_json_ld
  json_ld.to_h.compact.sort_by(&:first).to_h.to_json
end
```

spec/jekyll_seo_tag_integration_spec.rb
```diff
@@ -30,6 +30,10 @@
     expect(output).to match(%r!Jekyll v#{version}!i)
   end
 
+  it "outputs JSON sorted by key" do
+    expect(json.strip).to eql("{\"@context\":\"https://schema.org\",\"@type\":\"WebPage\",\"url\":\"/page.html\"}")
```
ashmaroli, line 34: How about using `%()` or outer single-quotes to avoid escaping the JSON's double-quotes?

```suggestion
    expect(json.strip).to eql('{"@context":"https://schema.org","@type":"WebPage","url":"/page.html"}')
```

ashmaroli (commented):
I recommend optimizing as much as possible to reduce impact on performance.

lib/jekyll-seo-tag/drop.rb
parkr: Since the `to_json` method is doing a bunch of work, it's probably best if I just sort there.

parkr:
@ashmaroli I figured we could make this a first-class thing and just change `to_json` to return a sorted object instead of adding a second method.

parkr:
Ok I decided to measure the `.sort_by(&:first).to_h` I'm adding. It's not great:

```text
Warming up --------------------------------------
            original    55.477k i/100ms
        with sort_by    30.221k i/100ms
Calculating -------------------------------------
            original    539.839k (± 8.8%) i/s -      2.718M in   5.078878s
        with sort_by    322.651k (± 2.8%) i/s -      1.632M in   5.062437s
```

I went up the chain and am taking advantage of the fact that Hashes preserve insertion order. So if you insert "@context" before "@type", then iterating over the Hash (and serializing it to JSON) will give you "@context" first, and "@type" second consistently. I overwrote the original Drop#to_h to sort the keys.

ashmaroli:
> taking advantage of the fact that Hashes preserve insertion order.. overwrote the original Drop#to_h to sort the keys.

This is a wonderful idea! Good job! :clap: 

**I think this is sufficient for an approval now**, but definitely could be optimized further..

Essentially, the `json_ld.to_json` call currently consists of numerous iterations:
```ruby
def to_json(state)
  keys.sort.each_with_object({}) do |(key, _), result|
    result[key] = self[key]
  end.compact.to_json(state)
end
```
:point_right: **`.sort.each_with_object.compact`**

I'll think about this till tomorrow. and shall register my approval then.
Will merge after Matt provides his inputs too.

Thank you very much @parkr

parkr:
@ashmaroli Great point! Further optimized to add an `if <value>` to the `each_with_object`. I think we need to keep `keys.sort` but a good optimization is to get rid of `compact` if not necessary!

ashmaroli:
:+1:

However, this PR ***may have exposed a bug in Jekyll Core***.. Specifically:
```ruby
keys.sort.each_with_object({}) do |(key, _), result|
```
From Jekyll Core codebase [`Drop#keys`](https://github.com/jekyll/jekyll/blob/d46bef774a8f91e7f58d982cccdface7a3d5d26b/lib/jekyll/drops/drop.rb#L192-L196) is **a flat array**:
```ruby
      def keys
        (content_methods |
          mutations.keys |
          fallback_data.keys).flatten
      end
```
Why is the `.each_with_object` block destructuring the **array** into pairs?
```ruby
keys.sort.each_with_object({}) do |(key, _), result|
```
instead of:
```ruby
keys.sort.each_with_object({}) do |key, result|
```

parkr:
Yeah I'm not sure why we did that. I'm going to leave it as-is to match Jekyll Core.

ashmaroli (commented):
One final edit and this is good to go.

ashmaroli (approved):

parkr:
@ashmaroli let's give Matt until the weekend. If he's not free to take a look by then, then I think your and my approval counts as 2 Jekyll maintainers' approval so we can merge and cut a release.

mattr- (approved):

ashmaroli:
Thanks @parkr
@jekyllbot:  merge +minor

parkr:
Thanks @mattr-! Thanks @ashmaroli!

PR #459, opened 2022-02-04
Add Ruby 3.1 to CI matrix

petergoldstein:
Adds Ruby 3.1 to the CI matrix.  Passes on my fork.

Also quotes 3.0 so it isn't truncated to 3.  Because of this truncation, an unquoted 3.0 loads the latest Ruby 3 - at this time 3.1.0 - as opposed to a 3.0.x version.  Adding quotes resolves this issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,8 @@ jobs:
         ruby_version:
         - 2.5
         - 2.7
-        - 3.0
+        - '3.0'
+        - 3.1
         jekyll_version:
         - "~> 4.0"
     steps:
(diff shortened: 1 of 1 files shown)


ashmaroli:
Thank you for your submission @petergoldstein.
We're not keen on testing with every supported Ruby version. (Notice how Ruby 2.6 is not in the list?)
I'd prefer having Ruby 3.0 replaced with 3.1 instead.

petergoldstein:
@ashmaroli Ok.  I've removed Ruby 3.0

ashmaroli (approved):

ashmaroli:
Thank you @petergoldstein
@jekyllbot: merge +dev

PR #464, opened 2022-07-12
Added the `twitter:description` tag

amirmasoudabdol:
I think we are missing the `twitter:description` tag, right?

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -18,6 +18,7 @@
 {% if seo_tag.description %}
   <meta name="description" content="{{ seo_tag.description }}" />
   <meta property="og:description" content="{{ seo_tag.description }}" />
+  <meta property="twitter:description" content="{{ seo_tag.description }}" />
 {% endif %}
 
 {% if site.url %}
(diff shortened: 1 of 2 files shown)


mattr-:
Does twitter not use the opengraph tag?

amirmasoudabdol:
I think it doesn't, but I cannot really be sure. I have noticed that if I don't include the `twitter:description` sometimes the description is missing when I share the article on Twitter! 🤷🏻‍♂️

mattr-:
Could you update the tests to include this new tag? You should be able to modify some already existing tests. Once that's done, this should be good to merge. Thanks!

amirmasoudabdol:
I have absolutely no idea how Ruby's tests work, but I think I did something! 😅

pathawks:
> When the Twitter card processor looks for tags on a page, it first checks for the Twitter-specific property, and if not present, falls back to the supported Open Graph property. This allows for both to be defined on the page independently, and minimizes the amount of duplicate markup required to describe content and experience.

https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started#opengraph

Reading the documentation, I assumed it would use the existing `og:description`. I wonder if there is some other reason why the description sometimes does not appear.

amirmasoudabdol:
I cannot really tell, but it should be fine if we provide an extra tag for twitter as well, right?

mattr-:
I think it's fine to include the twitter tag here. It would be nice if twitter would reliably use `og:description` but that doesn't seem to be the case. I'm also open to including other platform specific tags for large enough platforms.

mattr- (approved):
Nice work! 🎇

mattr-:
@jekyllbot: merge +minor

PR #472, opened 2022-10-18
Updated condition requiring tagline or description

uy-rrodriguez:
This PR updates the logic generating titles.

The title property was updated in #356 to prefer tagline instead of description. However, the condition on the `if` just above was not updated, so if `site.description` is not defined, then tagline won't be used.


imustafin:
I faced a problem with a website having `tagline` but no `description`. It might be confusing that adding a `description` makes `tagline` to appear somewhere.

Just wanted to make the same PR now. What is the status of this PR?

I guess @mattr- is active in this repo. Is it possible to merge this? Maybe @uy-rrodriguez could also update the branch, if not then I can make a similar PR.

PR #475, opened 2022-11-01
Makes template w3c valid

Koziolek:
All meta tags in template are void tags and should be open – end with `>` not `/>`. 

This change makes meta tags valid in w3c validator.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -3,123 +3,123 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
To my knowledge, the plugin doesn't differentiate between HTML and XHTML. Therefore, the "closing bits" are perhaps there for compatibility with XHTML documents / templates.

sasadangelo:
Discussed here:
https://github.com/jekyll/jekyll-seo-tag/pull/480

Author prefers to mantain the XHTML compatibility.

PR #481, opened 2023-01-22
Make the meta generator tag optional

danger89:
Webmasters are able to decide themselves whether they want to include a generate meta HTML tag or not using `seo_tag.skip_generator`.

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -3,7 +3,9 @@
   <title>{{ seo_tag.title }}</title>
 {% endif %}
 
+{% unless seo_tag.skip_generator %}
 <meta name="generator" content="Jekyll v{{ jekyll.version }}" />
+{% endif %}
 
 {% if seo_tag.page_title %}
   <meta property="og:title" content="{{ seo_tag.page_title }}" />


ashmaroli:
Prior art: #404

ashmaroli:
@danger89 We, the maintainers, have not reached at a consensus on whether the meta tag ought to be removed completely or rendered conditionally.
Regardless, I'd like to leave one comment as feedback to this pull request:
*How exactly are webmasters supposed to opt to render / not render the meta tag? They do not have control over the `seotag` object from their config file. You have not included any documentation / integration-test to demonstrate usage.

danger89:
https://github.com/jekyll/jekyll-seo-tag/pull/404

Is not apply review comments anymore. I applied the patch correctly.

danger89:
> @danger89 _How exactly are webmasters supposed to opt to render / not render the meta tag? They do not have control over the `seotag` object from their config file. You have not included any documentation / integration-test to demonstrate usage._

Well I think we have control over this. A bit similar to the title (which is also using the seo tag):

https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/advanced-usage.md#disabling-title-output

ashmaroli:
> A bit similar to the title (which is also using the seo tag).

@danger89 I'm aware of the connection of "title" with `{{ seo_tag }}`. But have you tested the proposed implementation at your end?

danger89:
I will test it soon. And also create the required integration tests as well as user documentation.

pegvin:
I have no clue why this isn't merged yet, it doesn't break anything, and it's a good feature that.

melroy89:
> I have no clue why this isn't merged yet, it doesn't break anything, and it's a good feature that.

Agreed. It's a simple but effective change.

pegvin:
I found a simple fix to this, I just switched to @lumeland, it's far better and faster than jekyll and community support is pretty alive too.

melroy89:
> I found a simple fix to this, I just switched to @lumeland, it's far better and faster than jekyll and community support is pretty alive too.

That is just a weird statement. Just because of this small missing optional meta tag generator? You have still full control over your plugins with Jekyll just like with Hugo. In all cases you can clone the code and made changes yourself, you are in full control how you want to use it or even extend it to your own liking. That being said, I do have they just purge these PRs, so I don't need to maintain my own fork.

PR #489, opened 2023-03-25
feat: support distinct seo image and page image

tsypuk:
Allows to have different or same images for rendering on the top page and for inside meta header when sharing link.

- If image.twitter or image.facebook is present than they will be used to generate og:image and twitter:image of SEO HEAD.
- Otherwise if only image or image:path present - than it will be present on both page image and seo tags.


tsypuk:
@ashmaroli please take a look at PR ^^

mattr-:
This PR is old and looking at the implementation, I'm not sure why it's important to have the order be this way. If this is still a change you're interested in, please re-submit. Thanks!

PR #492, opened 2023-06-30
Truncate the description value to 100 words

kemenaran:
Currently, on post pages, the value of the `description`, `og: description`, `twitter: description` and `application/ld-json` tags represent the whole text of the post.

Which means that the text of the post is included **five times** in the HTML document – adding to the page weight needlessly.

This commit ensures the description string is truncated to a sane value.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -80,6 +80,16 @@ You can set it the same way as the other author properties. For example, you can
     url: https://example.com/
   ```
 
+### Customizing description length
+
+By default the description is limited to the first 50 words of the full content.
+
+You can adjust this limit either at the page level or at the site level, by using the `seo_description_max_words` page property:
+
+```yml
+  seo_description_max_words: 100
+```
+
 ### Customizing JSON-LD output
 
 The following options can be set for any particular page. While the default options are meant to serve most users in the most common circumstances, there may be situations where more precise control is necessary.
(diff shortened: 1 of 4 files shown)


y377:
@kemenaran If it can be customized according to the percentage, this should be the ultimate solution: for example, from 10% to 90% of the full text (if the description is defined in the preface), it is recommended to consider it!

kemenaran:
@mattr- would you consider this improvement mergeable in the current state – or do you prefer to make it configurable?

mattr-:
280 characters feels like a twitter specific limitation. Regardless of that, a blanket truncation at 280 characters doesn't take into account word or sentence endings, and it feels like we should do that before this PR is mergeable.

kemenaran:
@mattr- thanks for your review! I updated the PR to:

- Limit the description length based on words count (instead of characters count);
- Make the maximum word count configurable (per page or site-wide).

I also added tests, and a new documentation section describing the behavior.

Does that look reasonable to you?

mattr- (approved):
Thanks for updating this! Everything here looks good to me!

kemenaran:
Yup, I just pushed fixes to the rubocop issues, it should be good now.

mattr-:
Thanks for tackling this!

@jekyllbot: merge +minor

y377:
@jekyllbot @kemenaran 
Thank you for your great efforts. Please tell me how to update the improved code. Because the plug-in has been installed before, no matter whether you use `gem install jekyll-seo-tag` or `gem update jekyll-seo-tag`, it will not help.

mattr-:
You'll need to wait on a plugin release or tell bundler to fetch the code from the `git` repo in order to use this new feature.

PR #505, opened 2024-04-11
feat: add modified_time meta tag

dimaslanjaka:
add modified date meta tag, useful for SEO

### DIFF
--- a/lib/template.html
+++ b/lib/template.html
@@ -46,6 +46,7 @@
 {% if page.date %}
   <meta property="og:type" content="article" />
   <meta property="article:published_time" content="{{ page.date | date_to_xmlschema }}" />
+  <meta property="article:modified_time" content="{{ page.updated | default: page.date | date_to_xmlschema }}"/>
 {% else %}
   <meta property="og:type" content="website" />
 {% endif %}


CookiePLMonster (changes requested):
This plugin, as well as `jekyll-feed`, make use of `page.last_modified_at`. It would make more sense to use this instead of `page.updated`. This way the SEO tag would match what `jekyll-feed` outputs:
https://github.com/jekyll/jekyll-feed/blob/43cc5d51e30b59ab8d8ba970e42c8c437e2b233f/lib/jekyll-feed/feed.xml#L78

@ashmaroli otherwise this looks good I think? Pinging you directly as per a comment in https://github.com/jekyll/jekyll-seo-tag/issues/512#issuecomment-2611289540 - despite not being the author of this PR, I wish to signal there is interest in this plugin.

ashmaroli:
I agree with @CookiePLMonster here. It would be better to use the `last_modified_at` attribute here for party between generated feed and SEO markup.

CookiePLMonster:
Furthermore, even in this plugin, `page.last_modified_at` is already used for LD-JSON generation:
https://github.com/jekyll/jekyll-seo-tag/blob/f449b1af64cbaf9080e7a3c677090cfe6e8f0021/spec/jekyll_seo_tag/drop_spec.rb#L298-L315

Example from my website:
`<script type="application/ld+json">
{"@context":"https://schema.org","@type":"BlogPosting","author":{"@type":"Person","name":"Silent"},"dateModified":"2025-04-23T13:30:00+00:00","datePublished":"2025-04-23T13:30:00+00:00"
```

CookiePLMonster:
@ashmaroli Looking at https://github.com/jekyll/jekyll-seo-tag/blob/f449b1af64cbaf9080e7a3c677090cfe6e8f0021/lib/jekyll-seo-tag/drop.rb#L114-L123, the current `article:published_time` isn't exactly great either and instead it should be `seo_tag.published_time`. I can prepare an alternative PR tomorrow correcting this + modification date.

CookiePLMonster:
Additionally, this duplicates #447, albeit using `last_modified_at`.

mattr-:
Closing in favor of #515 and this feature is now available in 2.9.0

PR #516, opened 2025-05-02
Bump versions in workflows

CookiePLMonster:
* Update actions/checkout
* Update the Ruby versions matrix to 2.5 (lowest supported) and all currently maintained Ruby versions (3.2-3.4)

I kept 2.5 in the build matrix as it's the lowest supported Ruby version in the `.gemspec` file. I'm not sure if it makes sense to keep 2.7 around, so I instead opted for the currently supported versions. For CI actions that deal with both Jekyll 3.x and 4.x, I opted to use Ruby 3.2 as the oldest currently maintained version, as Jekyll 3.x doesn't work with Ruby 3.4 due to the `base64` dependency no longer being part of the standard library. For tests, I wasn't sure what Ruby version to use so I left it at 2.5.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
       matrix:
         ruby_version:
         - 2.5
-        - 3.2
         - 3.3
         - 3.4
         jekyll_version:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,12 +20,13 @@ jobs:
       matrix:
         ruby_version:
         - 2.5
-        - 2.7
-        - 3.1
+        - 3.2
+        - 3.3
+        - 3.4
         jekyll_version:
         - "~> 4.0"
     steps:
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
Should the version matrix perhaps align with Jekyll 4.x https://github.com/jekyll/jekyll/blob/master/.github/workflows/ci.yml#L25-L30 + Ruby 2.5 as the lowest supported version?

CookiePLMonster:
If CI jobs are running against 2.7, should the minimum Ruby requirement for this gem be bumped too?

ashmaroli:
> If CI jobs are running against 2.7, should the minimum Ruby requirement for this gem be bumped too?

Not necessarily. The plugin can still be installed and run on Ruby 2.5 and Ruby 2.6 (albeit no longer tested with and therefore no longer recommended).
We bump the *required* Ruby-version when we introduce Ruby 2.7-specific syntax in `lib` files.

ashmaroli (approved):

PR #517, opened 2025-09-10
Add support for title_category

CodeSandwich:
This PR is best described by the diff of the changes to `docs/advanced-usage.md` it includes.

Personally I need it for a website that serves multiple longer texts divided into pages, like books divided into chapters. The currently generated `<title>` and title SEO tags just aren't enough, they are lacking the crucial indication that some pages are parts of larger entities, but smaller than the entire website. I also think that it will be useful for other structured texts like documentations where the context of each page is important.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -189,3 +189,24 @@ seo_paginator_message: "%<current>s / %<total>s | "
 
 While the value can be any string text, we recommend using a Ruby string-template containing the variables `current` and `total`
 similar to the example above, to incorporate the current page-number and total number of paginated pages in the title.
+
+### Adding a page title category
+
+You can optionally add a page category to its title.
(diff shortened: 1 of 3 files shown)


mattr- (commented):
This looks really good. I appreciate the documentation and test updates here as well. I left one suggestion. Let me know what you decide to do with the suggestion and I can proceed from there.

Thanks for contributing! ❤️

mattr-:
There will be a release at some point soon-ish. I'm in the middle of reworking the release processes for nearly all of our gems, so it might be awhile longer.

CookiePLMonster:
Hey, I'm just leaving this comment to express interest in this feature, so it would be great to see a new "official" release incorporating it.

Additionally, would it be possible to let layouts specify title categories? At the moment, I seem to only be able to specify them with config defaults.

PR #521, opened 2026-01-14
feat: Add canonical=false option to disable canonical URL output

rathboma:
## Summary

Adds a new `canonical=false` option to disable the `<link rel="canonical">` tag output, similar to the existing `title=false` option.

## Use Case

When using multilingual plugins like [jekyll-polyglot](https://github.com/untra/polyglot), the canonical URL needs special handling:
- Translated pages should have canonical URLs pointing to their language-specific URLs
- Fallback pages (using default language content) may need canonical URLs pointing to the default language version

By allowing users to disable the canonical output from jekyll-seo-tag, they can use their multilingual plugin's canonical URL generation instead.

## Usage

```liquid
{% seo canonical=false %}
```

This disables only the `<link rel="canonical">` tag while still outputting all other SEO tags including `<meta property="og:url">`.

## Changes

- Added `canonical?` method to `Drop` class (similar to `title?`)
- Updated template to conditionally output canonical tag
- Added unit tests for `canonical?` method
- Added integration tests for `canonical=false`
- Updated advanced-usage.md documentation

## Testing

All 222 existing tests pass, plus 5 new tests for the canonical feature.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -12,6 +12,18 @@ If for some reason, you don't want the plugin to output `<title>` tags on each p
 ```
 <!-- {% endraw %} -->
 
+### Disabling `<link rel="canonical">` output
+
+If you're using another plugin to generate canonical URLs (such as [jekyll-polyglot](https://github.com/untra/polyglot) for multilingual sites), you can disable the canonical URL output:
+
+<!-- {% raw %} -->
+```
+{% seo canonical=false %}
+```
+<!-- {% endraw %} -->
+
+This will prevent the plugin from outputting the `<link rel="canonical">` tag, while still outputting all other SEO tags including the `<meta property="og:url">` tag.
+
 ### Author information
 
 Author information is used to propagate the `creator` field of Twitter summary cards. This should be an author-specific, not site-wide Twitter handle (the site-wide username be stored as `site.twitter.username`).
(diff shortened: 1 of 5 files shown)


mattr-:
@rathboma this looks good. thanks for the contribution! If you could fix up the style guide issues, that would unblock merging this. I've pulled them out of the action to make it easier
```
lib/jekyll-seo-tag/drop.rb:38:16: C: Naming/MemoizedInstanceVariableName: Memoized variable @display_canonical does not match method name canonical?. Use @canonical instead.
        return @display_canonical if defined?(@display_canonical)
               ^^^^^^^^^^^^^^^^^^
lib/jekyll-seo-tag/drop.rb:38:47: C: Naming/MemoizedInstanceVariableName: Memoized variable @display_canonical does not match method name canonical?. Use @canonical instead.
        return @display_canonical if defined?(@display_canonical)
                                              ^^^^^^^^^^^^^^^^^^
lib/jekyll-seo-tag/drop.rb:40:9: C: Naming/MemoizedInstanceVariableName: Memoized variable @display_canonical does not match method name canonical?. Use @canonical instead.
        @display_canonical = (@text !~ %r!canonical=false!i)
```

rathboma:
@mattr- sorry for the delay! This should be good to go now

mattr- (approved):
LGTM!

@jekyllbot: merge +minor

Issue #503, opened 2024-03-03
How can I disable LD+JSON from included tags?


autogris:
I'm using specific tags for ld+json which requires a fine tuning, and would be nice to keep seo-tags just for the <meta>s but I would need to disable its ld+json feature. Is it possible, if so, how?

Issue #473, opened 2022-10-18
Critical feature req: Ability to turn off or better modify JSON-LD


maxnorton:
I'm finding this plugin very useful for its handling of og tags and Twitter cards but there are so few options to customize JSON-LD output that it is impossible to serve Google even a basic level of rich data. [Issue 135](https://github.com/jekyll/jekyll-seo-tag/issues/135) is related but appears to have been closed without any resolution. 

Would it not be straightforward to either (a) allow users to turn off JSON-LD and handle Schema markup themselves, or (b) enable more Schema properties? The discussion in the feature above is around products, but my use case is completely different (details about a person) and it is just as important to be able to customize the output.

tpluscode:
I second this. I'm adding SEO with the plugin to a site where I already created a complex JSON-LD markup, including opening hours etc.

I would ideally disable the generated output. Config like...

```yaml
seo:
  jsonld: false
```

maxnorton:
@tpluscode FWIW, I have a fork [here](https://github.com/maxnorton/jekyll-seo-tag) where I started trying to implement a liquid tag to turn it off. I didn't get it working, but the last commit simply comments out the JSON-LD injection in lib/template.html. I'm not planning on keeping the fork up to date but you're welcome to use it. If I get the time I will get the liquid tag working and open a PR.

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code>/<code>main</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

Issue #135, opened 2016-10-25
Make json-ld optional


queso:
The json-ld seems to have very limited options built in. I like this package/plugin and would love to keep using it, but I want to turn off the json-ld output if I need to. It seems https://github.com/jekyll/jekyll-seo-tag/blob/master/lib/template.html#L179 just rolls straight in and ouputs the code.

Perhaps we could just add an easy check to make sure page.seo exists before we output the code? That would allow me to set seo: false and add my own markup for json-ld, like for a @type of Product, etc.

queso:
For those reading this issue, it looks like Google will add data from microdata and json-ld, so I was able to add html markup to at least get an on page product working. Would still love a way to turn off json-ld so I can override with specific customizations down the road.

I like the other meta tags such as the social stuff, so I don't want to just pull out the seo tag all together, either.

DirtyF:
Could you provide some concrete examples, maybe we can enhance the current JSON-LD?

queso:
Sure, I guess so, this is all just part of google's documentation on JSON-LD...

Here is an example of json-ld for describing a product:
```
<script type="application/ld+json">
{
  "@context": "http://schema.org/",
  "@type": "Product",
  "name": "Executive Anvil",
  "image": "http://www.example.com/anvil_executive.jpg",
  "description": "Sleeker than ACME's Classic Anvil, the Executive Anvil is perfect for the business traveler looking for something to drop from a height.",
  "mpn": "925872",
  "brand": {
    "@type": "Thing",
    "name": "ACME"
  },
  "aggregateRating": {
    "@type": "AggregateRating",
    "ratingValue": "4.4",
    "reviewCount": "89"
  },
  "offers": {
    "@type": "Offer",
    "priceCurrency": "USD",
    "price": "119.99",
    "priceValidUntil": "2020-11-05",
    "itemCondition": "http://schema.org/UsedCondition",
    "availability": "http://schema.org/InStock",
    "seller": {
      "@type": "Organization",
      "name": "Executive Objects"
    }
  }
}
</script>
```

My goal is to have that outputted on this page, which has a book for sale on it: http://wendyowensbooks.com/books/burning-destiny. The idea is that Google can pick up the output for things like price, reviews, etc and show it in the search results.

queso:
[...]

PR #522, opened 2026-04-19
Allow disabling json-ld output

imustafin:
Closes or addresses #473, #135, #503.

Linked issues give background on why this might be a useful feature to have.

Personally I prefer to use Structured Date instead of JSON-LD since it is easier to configure using templates and it is still supported by Google and some other search engines.


mattr-:
Google recommends using JSON-LD for structured data so turning it off seems counterintuitive. I will happily review and merge PRs that make improvements to our json-ld support but I won't support turning it off at the moment.

simonoff:
yes, but instead of having ability to pass whole schema to its render super limited.

PR #523, opened 2026-04-19
Allow removing site title from page title

imustafin:
The default `homepage_or_about?` for titles does not work well with some non-default url templates.

For example, if the homepage has a language prefix like `/en/index.html`, it will not be considered a homepage, which will lead to site title being appended.

The proposed approach allows configuring this setting per page, giving flexibility when it is needed.

This PR also collects the title-related docs together in `advanced-usage.md`.


spec/jekyll_seo_tag/drop_spec.rb
```diff
@@ -168,6 +168,15 @@
...
       end
+
+      context "with a home_or_about page and a site title" do
+        let(:config) { { "title" => "site title" } }
+        let(:page_meta) { { "permalink" => "/", "title" => "page title" } }
+
+        it "is just the page title" do
+          expect(subject.title).to eql("page title")
+        end
+      end
```
mattr-: This doesn't seem to exercise the new code path for handling `page.key?("homepage_or_abort")` Could you add a test for that please?

imustafin: I clarified what I meant there. Now I have a spec for the normal homepage case (permalink is `/`) and the new override with `homepage_or_about`.

Sorry for the delay.

